PR #4, opened 2020-07-17
Expose the config for client side caches, and enable them

kotharironak:
- enables the client-side caches for most common types of ID-based queries
- adds new cache for queries based on identifying attributes and type
- exposes the cache-related constants to config

### DIFF
--- a/entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
+++ b/entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
@@ -32,13 +32,14 @@ public class EdsCacheClient implements EdsClient {
   private final EntityDataServiceClient client;
 
   EdsCacheClient(EntityServiceClientConfig entityServiceClientConfig) {
-    this(new EntityDataServiceClient(entityServiceClientConfig));
-    initCache(entityServiceClientConfig.getCacheConfig());
+    this(new EntityDataServiceClient(entityServiceClientConfig),
+        entityServiceClientConfig.getCacheConfig());
   }
 
-  public EdsCacheClient(EntityDataServiceClient client) {
+  public EdsCacheClient(EntityDataServiceClient client,
+      EntityServiceClientCacheConfig cacheConfig) {
     this.client = client;
-    initCache(EntityServiceClientCacheConfig.DEFAULT);
+    initCache(cacheConfig);
   }
 
   private void initCache(EntityServiceClientCacheConfig cacheConfig) {
--- a/entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
+++ b/entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
@@ -92,8 +92,8 @@ public class EdsCacheClient implements EdsClient {
     EdsTypeAndIdAttributesCacheKey key = new EdsTypeAndIdAttributesCacheKey(tenantId,
         byIdentifyingAttributes);
     try {
-      String entityId = entityIdsCache.get(key).orElse(null);
-      return (entityId != null) ? getById(tenantId, entityId) : null;
+      Optional<String> entityId = entityIdsCache.get(key);
+      return (entityId.isPresent()) ? getById(tenantId, entityId.get()) : null;
     } catch (ExecutionException e) {
       LOG.error("Failed to fetch entity of tenantId: {}, entityId: {}",
           key.tenantId, key.byTypeAndIdentifyingAttributes, e);
--- a/entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
+++ b/entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
@@ -10,6 +10,7 @@ import java.util.Set;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import javax.annotation.Nonnull;
+import org.hypertrace.entity.data.service.client.exception.NotFoundException;
 import org.hypertrace.entity.data.service.v1.ByTypeAndIdentifyingAttributes;
 import org.hypertrace.entity.data.service.v1.EnrichedEntities;
 import org.hypertrace.entity.data.service.v1.EnrichedEntity;
@@ -25,9 +26,9 @@ import org.slf4j.LoggerFactory;
 public class EdsCacheClient implements EdsClient {
 
   private static final Logger LOG = LoggerFactory.getLogger(EdsCacheClient.class);
-  private LoadingCache<EdsCacheKey, Optional<EnrichedEntity>> enrichedEntityCache;
-  private LoadingCache<EdsCacheKey, Optional<Entity>> entityCache;
-  private LoadingCache<EdsTypeAndIdAttributesCacheKey, Optional<String>> entityIdsCache;
+  private LoadingCache<EdsCacheKey, EnrichedEntity> enrichedEntityCache;
+  private LoadingCache<EdsCacheKey, Entity> entityCache;
+  private LoadingCache<EdsTypeAndIdAttributesCacheKey, String> entityIdsCache;
 
   private final EntityDataServiceClient client;
 
@@ -48,9 +49,13 @@ public class EdsCacheClient implements EdsClient {
             .expireAfterWrite(cacheConfig.getEnrichedEntityCacheExpiryMs(), TimeUnit.MILLISECONDS)
             .maximumSize(cacheConfig.getEnrichedEntityMaxCacheSize())
             .build(new CacheLoader<>() {
-              public Optional<EnrichedEntity> load(@Nonnull EdsCacheKey key) {
-                return Optional
-                    .ofNullable(client.getEnrichedEntityById(key.tenantId, key.entityId));
+              public EnrichedEntity load(@Nonnull EdsCacheKey key) throws Exception {
+                EnrichedEntity enrichedEntity = client
+                    .getEnrichedEntityById(key.tenantId, key.entityId);
+                if (enrichedEntity == null) {
+                  throw new NotFoundException("Enriched entity not found");
+                }
+                return enrichedEntity;
               }
             });
 
@@ -59,8 +64,12 @@ public class EdsCacheClient implements EdsClient {
             .expireAfterWrite(cacheConfig.getEntityCacheExpiryMs(), TimeUnit.MILLISECONDS)
             .maximumSize(cacheConfig.getEntityMaxCacheSize())
             .build(new CacheLoader<>() {
-              public Optional<Entity> load(@Nonnull EdsCacheKey key) {
-                return Optional.ofNullable(client.getById(key.tenantId, key.entityId));
+              public Entity load(@Nonnull EdsCacheKey key) throws Exception {
+                Entity entity = client.getById(key.tenantId, key.entityId);
+                if (entity == null) {
+                  throw new NotFoundException("Entity not found");
+                }
+                return entity;
               }
             });
 
@@ -68,16 +77,14 @@ public class EdsCacheClient implements EdsClient {
         .expireAfterWrite(cacheConfig.getEntityIdsCacheExpiryMs(), TimeUnit.MILLISECONDS)
         .maximumSize(cacheConfig.getEntityIdsMaxCacheSize())
         .build(new CacheLoader<>() {
-          public Optional<String> load(@Nonnull EdsTypeAndIdAttributesCacheKey key) {
+          public String load(@Nonnull EdsTypeAndIdAttributesCacheKey key) throws Exception {
             Entity entity = client.getByTypeAndIdentifyingAttributes(key.tenantId,
                 key.byTypeAndIdentifyingAttributes);
-            String entityId = null;
-            if (entity != null) {
-              entityId = entity.getEntityId();
-              entityCache.put(new EdsCacheKey(entity.getTenantId(), entity.getEntityId()),
-                  Optional.of(entity));
+            if (entity == null) {
+              throw new NotFoundException("Entity not found!!");
             }
-            return Optional.ofNullable(entityId);
+            entityCache.put(new EdsCacheKey(entity.getTenantId(), entity.getEntityId()), entity);
+            return entity.getEntityId();
           }
         });
   }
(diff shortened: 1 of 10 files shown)


avinashkolluru:
Looks good. I know we are fighting with time here, but is there any way to add a quick integration test to confirm its working as expected?

kotharironak:
@avinashkolluru existing integration tests should pass. And, will add specifically for getByTypeAndIdentifyingAttributes.

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
```diff
@@ -45,12 +54,31 @@ public EdsCacheClient(EntityDataServiceClient client) {
...
+        .expireAfterWrite(cacheConfig.getEntityIdsCacheExpiryMs(), TimeUnit.MILLISECONDS)
+        .maximumSize(cacheConfig.getEntityIdsMaxCacheSize())
+        .build(new CacheLoader<>() {
+          public Optional<String> load(@Nonnull EdsTypeAndIdAttributesCacheKey key) {
+            Entity entity = client.getByTypeAndIdentifyingAttributes(key.tenantId,
+                key.byTypeAndIdentifyingAttributes);
+            String entityId = null;
+            if (entity != null) {
+              entityId = entity.getEntityId();
+              entityCache.put(new EdsCacheKey(entity.getTenantId(), entity.getEntityId()),
```
buchi-busireddy, line 76: So, you're updating `entityCache` in addition to this cache when there is a hit?

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
```diff
@@ -61,7 +89,16 @@ public Entity upsert(Entity entity) {
   @Override
   public Entity getByTypeAndIdentifyingAttributes(String tenantId,
       ByTypeAndIdentifyingAttributes byIdentifyingAttributes) {
-    return client.getByTypeAndIdentifyingAttributes(tenantId, byIdentifyingAttributes);
+    EdsTypeAndIdAttributesCacheKey key = new EdsTypeAndIdAttributesCacheKey(tenantId,
+        byIdentifyingAttributes);
+    try {
+      String entityId = entityIdsCache.get(key).orElse(null);
```
buchi-busireddy, line 95: You shouldn't need `orElse` because the cache returns `Optional`

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheKey.java
```diff
@@ -14,4 +16,18 @@ public EdsCacheKey(String tenantId, String entityId) {
   public String toString() {
     return String.format("__%s__%s", this.tenantId, this.entityId);
   }
+
+  @Override
+  public int hashCode() {
```
buchi-busireddy, line 21: oops.. So, was this making the entire caches ineffective?

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsTypeAndIdAttributesCacheKey.java
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  @Override
+  public String toString() {
+    return String.format("__%s__%s", this.tenantId, this.byTypeAndIdentifyingAttributes);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(tenantId, byTypeAndIdentifyingAttributes);
```
buchi-busireddy, line 24: This triggers the `hashCode` calculation on the `byTypeAndIdentifyingAttributes` object internally. I've not fully seen the proto auto-generated code for that but did you think about performance impact due to that? Is it better to handle this on our own rather?

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsTypeAndIdAttributesCacheKey.java
```diff
@@ -0,0 +1,40 @@
...
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    EdsTypeAndIdAttributesCacheKey that = (EdsTypeAndIdAttributesCacheKey) o;
+    return Objects.equals(tenantId, that.tenantId)
+        && Objects.equals(byTypeAndIdentifyingAttributes, that.byTypeAndIdentifyingAttributes);
```
buchi-busireddy, line 37: Similarly equals comparison here too.

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheKey.java
avinashkolluru: Yes :)

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsTypeAndIdAttributesCacheKey.java
avinashkolluru: Might be tricky to maintain if we have a custom implementation

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
```diff
@@ -17,26 +17,35 @@
...
-  private final LoadingCache<EdsCacheKey, Optional<Entity>> entityCache;
+  private LoadingCache<EdsCacheKey, Optional<EnrichedEntity>> enrichedEntityCache;
+  private LoadingCache<EdsCacheKey, Optional<Entity>> entityCache;
+  private LoadingCache<EdsTypeAndIdAttributesCacheKey, Optional<String>> entityIdsCache;
 
   private final EntityDataServiceClient client;
 
   EdsCacheClient(EntityServiceClientConfig entityServiceClientConfig) {
     this(new EntityDataServiceClient(entityServiceClientConfig));
+    initCache(entityServiceClientConfig.getCacheConfig());
```
buchi-busireddy, line 36: This will initialize the caches twice because the `this()` call would have initialized with DEFAULT values already.
How about passing the cacheConfig as second param to the other constructor.

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
```diff
@@ -17,26 +17,35 @@
...
   }
 
   public EdsCacheClient(EntityDataServiceClient client) {
```
buchi-busireddy, line 39: When the client application has a EDSClient that it wants to reuse, it should pass the cache configs too. It can pass empty `Config` to use defaults though.

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsTypeAndIdAttributesCacheKey.java
```diff
@@ -0,0 +1,40 @@
+package org.hypertrace.entity.data.service.client;
+
+import java.util.Objects;
+import org.hypertrace.entity.data.service.v1.ByTypeAndIdentifyingAttributes;
+
+public class EdsTypeAndIdAttributesCacheKey {
```
buchi-busireddy, line 6: Unit test to verify the equals of different objects?

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
```diff
@@ -17,26 +17,35 @@
 import org.hypertrace.entity.data.service.v1.EntityRelationship;
 import org.hypertrace.entity.data.service.v1.EntityRelationships;
 import org.hypertrace.entity.data.service.v1.Query;
+import org.hypertrace.entity.service.client.config.EntityServiceClientCacheConfig;
 import org.hypertrace.entity.service.client.config.EntityServiceClientConfig;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class EdsCacheClient implements EdsClient {
```
buchi-busireddy, line 25: We should add unit tests to this class to unit test that the caching is effective.

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
kotharironak: It will be handle as any other entry. The property `expireAfterWriteNanos` is the cache level and its apply to all entries.

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
kotharironak: Done

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsTypeAndIdAttributesCacheKey.java
kotharironak: Scanned the internal Impl of generated proto's equals and hashCode. I think we should be fine.

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
```diff
@@ -45,12 +54,31 @@ public EdsCacheClient(EntityDataServiceClient client) {
...
+
+    this.entityIdsCache = CacheBuilder.newBuilder()
+        .expireAfterWrite(cacheConfig.getEntityIdsCacheExpiryMs(), TimeUnit.MILLISECONDS)
+        .maximumSize(cacheConfig.getEntityIdsMaxCacheSize())
+        .build(new CacheLoader<>() {
+          public Optional<String> load(@Nonnull EdsTypeAndIdAttributesCacheKey key) {
+            Entity entity = client.getByTypeAndIdentifyingAttributes(key.tenantId,
+                key.byTypeAndIdentifyingAttributes);
+            String entityId = null;
+            if (entity != null) {
```
inespo, line 74: Do you want to cache null as Optional.absent() or not cache at all?

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
kotharironak: Hmm. The other one is also public interface `public EdsCacheClient(EntityDataServiceClient client)`, and there was one consumer that directly uses this. So, I had to override this way.

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
kotharironak: See the previous comment.

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
inespo: Why do you need to explicitly add .put for LoadingCache? This could cause race-condition. It is better to leverage the cache's mechanism and call .invalidate. The next time a consumer calls "get" it will load from DB.

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
buchi-busireddy: You can either add one more constructor or change the public constructor too. Consumer will change in next version upgrade.

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
buchi-busireddy: The former. If you don't cache, you wouldn't even know that we queried for that entry right?

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsTypeAndIdAttributesCacheKey.java
kotharironak: Done.

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
kotharironak: We already have entity cache which is based on EdsCacheKey. We need one more based on identifying attributes. So, to avoid duplicates, we have added additional cache to store only entityIds (not full Entity) which are loaded via function - `getByTypeAndIdentifyingAttributes`. And, to avoid additional functional call back to DB which was done `getById` inside `getByTypeAndIdentifyingAttributes`, we are updating entity cache. LocalCache is already synchronized using a lock. So, I think, this should be fine.

avinashkolluru: We are putting into a different cache. Not sure what you mean by invalidate helping

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
kotharironak: Done.

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
kotharironak: Added basic test.

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
inespo: I see. Yes, in this case invalidate wouldn't help and as for the race-condition, you're not computing before putting so that wouldn't lead to the problem.

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
kotharironak: done

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
```diff
@@ -10,47 +10,83 @@
...
+        CacheBuilder.newBuilder()
+            .expireAfterWrite(cacheConfig.getEntityCacheExpiryMs(), TimeUnit.MILLISECONDS)
+            .maximumSize(cacheConfig.getEntityMaxCacheSize())
             .build(new CacheLoader<>() {
-              public Optional<Entity> load(@Nonnull EdsCacheKey key) {
-                return Optional.ofNullable(client.getById(key.tenantId, key.entityId));
+              public Entity load(@Nonnull EdsCacheKey key) throws Exception {
+                Entity entity = client.getById(key.tenantId, key.entityId);
+                if (entity == null) {
+                  throw new NotFoundException("Entity not found");
```
surajpuvvada, line 70: when can this happen ?

inespo: It's possible that gateway service passes an invalid entityID from public Graphql query or Explorer API.

surajpuvvada: got it

PR #5, opened 2020-07-24
Update grpc utils with changes for binary headers

mohit-a21:
Relevant version is 0.1.2

PR #6, opened 2020-07-24
Update logging level to warn for eds cache client

kotharironak:
This is not an erroneous scenario, so converting logging messages as a warning.


kotharironak:
PR - https://github.com/hypertrace/entity-service/pull/5 - will take care of snyk issue.

PR #10, opened 2020-08-03
Entity serivce changes to move configs to config-bootstrapper docker …

samarth-gupta-traceable:
Entity serivce changes to move configs to config-bootstrapper docker image


helm/values.yaml
```diff
@@ -126,5 +95,4 @@ logConfig:
 config-bootstrapper:
   job:
     prefix: entity
-
-configurationCommands: {}
+  configurationCommands: {}
```
kotharironak, line 98: is this an indentation issue?

samarth-gupta-traceable: @kotharironak  , no its intentional. `configurationCommands` value now needs to be passed to config-boostrapper helm chart.

PR #12, opened 2020-08-07
provides a way to override the mongo_host via an environment variable

kotharironak:


entity-service/src/main/resources/configs/common/application.conf
```diff
@@ -7,6 +7,7 @@ entity.service.config = {
     dataStoreType = mongo
     mongo {
       host = localhost
+      host = ${?MONGO_HOST}
       port = 27017
```
ravisingal, line 11: just wondering, should we have env variable for mongo port as well?

kotharironak: Had the same thought, but till now, we didn't have to override it, so, should we?

ravisingal: It's ok for now. We can add it later if required.

PR #22, opened 2020-09-08
Reducing log level to debug for miss hit in cache

kotharironak:


entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
```diff
@@ -101,8 +101,10 @@ public Entity getByTypeAndIdentifyingAttributes(String tenantId,
     try {
       return getById(tenantId, entityIdsCache.get(key));
     } catch (ExecutionException e) {
-      LOG.warn("Failed to fetch entity of tenantId: {}, entityId: {}",
-          key.tenantId, key.byTypeAndIdentifyingAttributes, e);
+      if (LOG.isDebugEnabled()) {
+        LOG.debug("Failed to fetch entity of tenantId: {}, entityId: {}",
+                key.tenantId, key.byTypeAndIdentifyingAttributes, e);
```
avinashkolluru, line 106:
```suggestion
                key.tenantId, key.byTypeAndIdentifyingAttributes);
```

avinashkolluru: I think stacktrace isn't printed for debug logs. So better get rid of that argument.

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
```diff
@@ -101,8 +101,10 @@ public Entity getByTypeAndIdentifyingAttributes(String tenantId,
     try {
       return getById(tenantId, entityIdsCache.get(key));
     } catch (ExecutionException e) {
-      LOG.warn("Failed to fetch entity of tenantId: {}, entityId: {}",
-          key.tenantId, key.byTypeAndIdentifyingAttributes, e);
+      if (LOG.isDebugEnabled()) {
```
jcchavezs, line 104: Not a Java expert but I am not 100% sure why we need to check if debug is enabled or not, can't we just do the call and since we are using the interpolation wildcards `{}` the serialization won't happen unless debug is truly enabled? cc @anuraaga

avinashkolluru: In this above example it is not needed. But lets say the argument is something like `key.tenantId + key.byTypeAndIdentifyingAttributes`
In such a case we are unnecessarily paying the cost of concatenation even though debug logs are not enabled. A simple concatenation may not be expensive, but that was just an example.

Ideally the check should be done for all log levels, but as a practice usually INFO logs are enabled by default and hence the check can be ignored(since it would evaluate to true)

avinashkolluru: So @kotharironak I think you can get rid of the explicit isDebugEnabled check above since the arguments involve no computation

jcchavezs: > In this above example it is not needed. But lets say the argument is something like key.tenantId + key.byTypeAndIdentifyingAttributes or key.toString()

I think in such case we could do something like `LOG.debug("Failed to fetch entity of concat: {}{}, entityId: {}", key.tenantId, key.byTypeAndIdentifyingAttributes, key)` but I totally agree on the part where a expensive operation takes place and we can't use wildcards then makes sense to use the if. In this case we don't tho.

avinashkolluru: Yes. The string concat was just an hypothetical example

kotharironak: The blow interface which takes more than 3 args,
```
public void debug(String format, Object... arguments)
``
1.  incur a little cost
```
<p>This form avoids superfluous string concatenation when the logger
     * is disabled for the DEBUG level. However, this variant incurs the hidden
     * (and relatively small) cost of creating an <code>Object[]</code> before invoking the method,
     * even if this logger is disabled for DEBUG. The variants taking
```
2. It also supports `throwable`, and we were passing ex

So, Let me remove `ex` argument and move to another method, and will remove isDebugEnabled.

kotharironak: @avinashkolluru @jcchavezs addressed comments.

avinashkolluru (approved):

PR #26, opened 2020-09-28
fix: convert literal const to attribute value during update

tim-mwangi:


entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -219,7 +219,10 @@ private void doUpdate(EntityUpdateRequest request, Map<String, String> attribute
         throw new IllegalArgumentException("Unknown attribute FQN " + attributeFqn);
       }
       String subDocPath = attributeFqnMap.get(attributeFqn);
-      String jsonValue = DocStoreJsonFormat.printer().print(setAttribute.getValue());
+      // Convert setAttribute LiteralConstant to AttributeValue. Need to be able to store an array
+      // literal constant as an array
+      AttributeValue.Builder attributeValueBuilder = EntityQueryConverter.convertToAttributeValue(setAttribute.getValue());
+      String jsonValue = DocStoreJsonFormat.printer().print(attributeValueBuilder);
```
aaron-steinfeld, line 225: Does the builder print in the same way as a built object? Seems ambiguous at a minimum, so may want to build on the line above.

tim-mwangi: We probably should. Was avoiding the extra build step since the method accepts both a builder and the built object.

aaron-steinfeld (approved):

aaron-steinfeld (approved):

ryanericson (approved):

aaron-steinfeld (approved):

buchi-busireddy (approved):

PR #28, opened 2020-10-07
fix: add entity labels caching client

tim-mwangi:
Still WIP.
- Need to implement an async refresh in the cache after write.
- Add unit tests

All Done. Ready for review. Also, github doesn't have a strikethrough?

### DIFF
--- a/entity-service-client/src/main/java/org/hypertrace/entity/query/service/client/EntityCacheKey.java
+++ b/entity-service-client/src/main/java/org/hypertrace/entity/query/service/client/EntityCacheKey.java
@@ -1,0 +1,39 @@
+package org.hypertrace.entity.query.service.client;
+
+import java.util.Map;
+import java.util.Objects;
+
+public class EntityCacheKey<T> {
+  private final T dataKey;
+  private final String tenantId;
+  private final Map<String, String> headers;
+
+  public EntityCacheKey(T key, String tenantId, Map<String, String> headers) {
+    this.dataKey = key;
+    this.tenantId = tenantId;
+    this.headers = headers;
+  }
+
+  public T getDataKey() {
+    return dataKey;
+  }
+
+  public Map<String, String> getHeaders() {
+    return headers;
+  }
+
+  // Only dataKey and tenantId are considered as part of the cache key. Headers are not.
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) return true;
+    if (o == null || getClass() != o.getClass()) return false;
+    EntityCacheKey that = (EntityCacheKey) o;
+    return Objects.equals(dataKey, that.dataKey) &&
+        Objects.equals(tenantId, that.tenantId);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(dataKey, tenantId);
+  }
+}
--- a/entity-service-client/src/main/java/org/hypertrace/entity/query/service/client/EntityLabelsCachingClient.java
+++ b/entity-service-client/src/main/java/org/hypertrace/entity/query/service/client/EntityLabelsCachingClient.java
@@ -27,10 +27,11 @@ import org.hypertrace.entity.query.service.v1.ResultSetChunk;
 import org.hypertrace.entity.query.service.v1.Row;
 import org.hypertrace.entity.query.service.v1.Value;
 import org.hypertrace.entity.query.service.v1.ValueType;
-import org.hypertrace.entity.service.client.config.EntityServiceClientConfig;
 
 public class EntityLabelsCachingClient implements EntityLabelsClient {
   private static final int THREAD_COUNT = 20;
+  private static final int CACHE_INITIAL_CAPACITY = 200;
+
   private static final ExecutorService executorService = Executors.newFixedThreadPool(THREAD_COUNT);
   private static final ListeningExecutorService listeningExecutorService = MoreExecutors.listeningDecorator(executorService);
 
@@ -53,20 +54,16 @@ public class EntityLabelsCachingClient implements EntityLabelsClient {
     }
   };
 
-  // TODO: Make these builder parameters configurable and move this to the constructor?
-  private final LoadingCache<EntityCacheKey<EntityTypeAndId>, List<String>> entityIdsToLabelsCache =
-      CacheBuilder.newBuilder()
-          .initialCapacity(500)
-          .maximumSize(10000)
-          .expireAfterWrite(10, TimeUnit.MINUTES)
-          .build(loader);
+  private final LoadingCache<EntityCacheKey<EntityTypeAndId>, List<String>> entityIdsToLabelsCache;
 
-  public EntityLabelsCachingClient(EntityServiceClientConfig entityServiceClientConfig) {
-    this.entityQueryServiceClient = new EntityQueryServiceClient(entityServiceClientConfig);
-  }
-
-  public EntityLabelsCachingClient(EntityQueryServiceClient entityQueryServiceClient) {
+  public EntityLabelsCachingClient(EntityQueryServiceClient entityQueryServiceClient,
+                                   EntityLabelsCachingClientConfig entityLabelsCachingClientConfig) {
     this.entityQueryServiceClient = entityQueryServiceClient;
+    this.entityIdsToLabelsCache = CacheBuilder.newBuilder()
+        .initialCapacity(CACHE_INITIAL_CAPACITY)
+        .maximumSize(entityLabelsCachingClientConfig.getMaximumSize())
+        .expireAfterWrite(entityLabelsCachingClientConfig.getExpiryTimeInMinutes(), TimeUnit.MINUTES)
+        .build(loader);
   }
 
   @Override
--- a/entity-service-client/src/main/java/org/hypertrace/entity/query/service/client/EntityLabelsCachingClient.java
+++ b/entity-service-client/src/main/java/org/hypertrace/entity/query/service/client/EntityLabelsCachingClient.java
@@ -1,0 +1,186 @@
+package org.hypertrace.entity.query.service.client;
+
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+import org.hypertrace.entity.query.service.v1.ColumnIdentifier;
+import org.hypertrace.entity.query.service.v1.EntityQueryRequest;
+import org.hypertrace.entity.query.service.v1.Expression;
+import org.hypertrace.entity.query.service.v1.Filter;
+import org.hypertrace.entity.query.service.v1.LiteralConstant;
+import org.hypertrace.entity.query.service.v1.Operator;
+import org.hypertrace.entity.query.service.v1.ResultSetChunk;
+import org.hypertrace.entity.query.service.v1.Row;
+import org.hypertrace.entity.query.service.v1.Value;
+import org.hypertrace.entity.query.service.v1.ValueType;
+import org.hypertrace.entity.service.client.config.EntityServiceClientConfig;
+
+public class EntityLabelsCachingClient implements EntityLabelsClient {
+  public EntityQueryServiceClient entityQueryServiceClient;
+
+  private final CacheLoader<EntityCacheKey<EntityTypeAndId>, List<String>> loader = new CacheLoader<>() {
+    @Override
+    public List<String> load(EntityCacheKey<EntityTypeAndId> key) {
+      return loadEntityLabelsForKeys3(List.of(key)).get(key);
+    }
+
+    @Override
+    public Map<EntityCacheKey<EntityTypeAndId>, List<String>> loadAll(Iterable<? extends EntityCacheKey<EntityTypeAndId>> keys) {
+      return loadEntityLabelsForKeys3(keys);
+    }
+  };
+
+  // TODO: Make these builder parameters configurable and move this to the constructor?
+  private final LoadingCache<EntityCacheKey<EntityTypeAndId>, List<String>> entityIdsToLabelsCache =
+      CacheBuilder.newBuilder()
+          .initialCapacity(500)
+          .maximumSize(10000)
+          .expireAfterWrite(10, TimeUnit.MINUTES)
+          .build(loader);
+
+  public EntityLabelsCachingClient(EntityServiceClientConfig entityServiceClientConfig) {
+    this.entityQueryServiceClient = new EntityQueryServiceClient(entityServiceClientConfig);
+  }
+
+  private static Expression.Builder createColumnExpression(String columnName) {
+    return Expression.newBuilder()
+        .setColumnIdentifier(ColumnIdentifier.newBuilder().setColumnName(columnName));
+  }
+
+  private static Expression.Builder createStringArrayLiteralExpression(List<String> strList) {
+    return Expression.newBuilder().setLiteral(
+        LiteralConstant.newBuilder().setValue(
+            Value.newBuilder()
+                .setValueType(ValueType.STRING_ARRAY)
+                .addAllStringArray(strList)
+        )
+    );
+  }
+
+  @Override
+  public List<String> getEntityLabelsForEntity(String idColumnName,
+                                               String labelsColumnName,
+                                               String id,
+                                               String type,
+                                               Map<String, String> headers,
+                                               String tenantId) {
+    try {
+      return entityIdsToLabelsCache.get(
+          new EntityCacheKey<>(new EntityTypeAndId(type, id, idColumnName, labelsColumnName),
+              tenantId,
+              headers
+          )
+      );
+    } catch (ExecutionException ex) {
+      throw new RuntimeException("EntityLabelsCachingClient exception while getting labels for an entity", ex);
+    }
+  }
+
+  @Override
+  public Map<String, List<String>> getEntityLabelsForEntities(String idColumnName,
+                                                              String labelsColumnName,
+                                                              List<String> ids,
+                                                              String type,
+                                                              Map<String, String> headers,
+                                                              String tenantId) {
+    List<EntityCacheKey<EntityTypeAndId>> entityCacheKeys = ids.stream()
+        .map(entityId -> new EntityTypeAndId(type, entityId, idColumnName, labelsColumnName))
+        .map(entityTypeAndId -> new EntityCacheKey<>(entityTypeAndId, tenantId, headers))
+        .collect(Collectors.toList());
+    try {
+      Map<EntityCacheKey<EntityTypeAndId>, List<String>> entityCacheKeyToLabelsMap =  entityIdsToLabelsCache.getAll(entityCacheKeys);
+      return entityCacheKeyToLabelsMap.entrySet().stream()
+          .collect(
+              Collectors.toMap(entry -> entry.getKey().getDataKey().getId(), Entry::getValue)
+          );
+    } catch (ExecutionException ex) {
+      throw new RuntimeException("EntityLabelsCachingClient exception while getting labels for entities", ex);
+    }
+  }
+
+  private Map<EntityCacheKey<EntityTypeAndId>, List<String>> loadEntityLabelsForKeys3(Iterable<? extends EntityCacheKey<EntityTypeAndId>> keys) {
+    Iterator<? extends EntityCacheKey<EntityTypeAndId>> iter = keys.iterator();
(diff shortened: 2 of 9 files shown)


entity-service-client/src/main/java/org/hypertrace/entity/query/service/client/EntityLabelsClient.java
```diff
@@ -0,0 +1,27 @@
+package org.hypertrace.entity.query.service.client;
+
+import java.util.List;
+import java.util.Map;
+
+public interface EntityLabelsClient {
+  List<String> getEntityLabelsForEntity(String idColumnName,
```
aaron-steinfeld, line 7: I thought you wanted things to be async?!

Regardless, I find this api a bit confusing. Why does the caller have to provide column names? And why are we explicitly unwrapping things like tenantIds and headers - that defeats the purpose of using a context which automatically propagates them.

Also, what's the reasoning to make this explicit to labels vs entities in general?

entity-service-client/src/main/java/org/hypertrace/entity/query/service/client/EntityLabelsClient.java
```diff
@@ -0,0 +1,27 @@
...
+                                        String tenantId);
+
+  Map<String, List<String>> getEntityLabelsForEntities(String idColumnName,
+                                                       String labelsColumnName,
+                                                       List<String> ids,
+                                                       String type,
+                                                       Map<String, String> headers,
+                                                       String tenantId);
+
+  List<String> getEntitiesWithLabels(String idColumnName,
```
aaron-steinfeld, line 21: This sounds like it should be returning entities - I'm guessing these are IDs?

aaron-steinfeld (commented):
haven't yet reviewed implementation - would like to drill into the api a bit first.

entity-service-client/src/main/java/org/hypertrace/entity/query/service/client/EntityLabelsClient.java
tim-mwangi: Yes. These are IDs. Will rename

entity-service-client/src/main/java/org/hypertrace/entity/query/service/client/EntityLabelsClient.java
tim-mwangi: RequestContext is defined in gateway-service that's why we have to pass in tenantId(needed for the cache key) and headers(needed by the actual entity-service client).

The caller provides column names since they are needed to construct the query to entity service in case of a cache miss. And yes, this cache is special for entity labels. In order to be able to make queries to entity service incase of a cache miss, we need all these arguments. Something more generic would need more thought and time :)

Also while I would like this to be async, it does not help at the moment since I need to know the entityIds whose labels I need to fetch. So might as well tackle that while making all API calls async.

entity-service-client/src/main/java/org/hypertrace/entity/query/service/client/EntityCacheKey.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.util.Map;
+import java.util.Objects;
+
+public class EntityCacheKey<T> {
+  private final T dataKey;
+  private final String tenantId;
+  private final Map<String, String> headers;
+
+  public EntityCacheKey(T key, String tenantId, Map<String, String> headers) {
```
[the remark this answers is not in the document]
tim-mwangi: RequestContext is defined in gateway-service.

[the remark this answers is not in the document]
aaron-steinfeld: I meant this one https://github.com/hypertrace/java-grpc-utils/blob/main/grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/RequestContext.java - that's generally what we're starting with right?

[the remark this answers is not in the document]
tim-mwangi: Yes, but we unpack it in GatewayServiceImpl as soon as we start handling the request. Also the grpc client takes in either tenantId or headers https://github.com/hypertrace/java-grpc-utils/blob/b16a4025066defb2cf4a5f8822c3a4f428e9f7cd/grpc-client-utils/src/main/java/org/hypertrace/core/grpcutils/client/GrpcClientRequestContextUtil.java#L40 . It creates a new grpc RequestContext object from these headers or the tenantId, different from the one that came into the server. To change this, will involve loads of changes.

entity-service-client/src/main/java/org/hypertrace/entity/query/service/client/EntityLabelsClient.java
aaron-steinfeld: >  In order to be able to make queries to entity service incase of a cache miss, we need all these arguments. 

Entity service exposes column names?!

> Something more generic would need more thought and time :)

That's fair - I'm not necessarily asking for changes yet, just trying to understand what led us here

I guess a more macro level question is - why is any of this needed? Gateway today has an entity fetcher (no idea if it caches, but presumably it could be added). What's special about labels that requires there own client when other entity-based attributes do not?

entity-service-client/src/main/java/org/hypertrace/entity/query/service/client/EntityCacheKey.java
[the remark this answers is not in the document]
aaron-steinfeld: So I don't really understand why we chose to do it that way. We don't have to change it, but we can easily go around it I think. For example for the Rx stuff:
https://github.com/hypertrace/java-grpc-utils/blob/b16a4025066defb2cf4a5f8822c3a4f428e9f7cd/grpc-client-rx-utils/src/main/java/org/hypertrace/core/grpcutils/client/rx/DefaultGrpcRxExecutionContext.java

I did this as a wrapper rather than putting it directly on request context so as not to introduce an Rx dependency across the platform, but other than that - we're just creating a new GRPC context, setting our own context as a key'd value in it, then using the grpc context's built in methods to make the function call. There's no need to build up a new context explicitly - we could just add `call`/`run`/`stream` methods to `RequestContext`.

That said, if it makes sense to do it the new way we should deprecate and transition the old stuff, but that can be tech debt.

entity-service-client/src/main/java/org/hypertrace/entity/query/service/client/EntityLabelsClient.java
tim-mwangi: > Entity service exposes column names?!

No. They are based on attribute service. So they correspond to the attribute metadata ids. It's expected that entity-service respects that but we do pass them into the query as strings.

> That's fair - I'm not necessarily asking for changes yet, just trying to understand what led us here
> 
> I guess a more macro level question is - why is any of this needed? Gateway today has an entity fetcher (no idea if it caches, but presumably it could be added). What's special about labels that requires there own client when other entity-based attributes do not?

Oh. 2 main motivations:
- Performance considerations; avoiding queries into 2 data sources like we were doing before since the labels will be used to annotate the entities - Service and API - wherever they appear on the UI.
- Tied to the bullet above, since we will be associating entity labels as foreign columns in other assets/scopes, and we are not writing these to Pinot, then we need a performant way to append labels to results of these other scopes.

So, this is a quick cache to get the feature out there and maintain some of the performance numbers that we have. @sanjay-nagaraj and I were talking about exploring a real cache service like redis.

entity-service-client/src/main/java/org/hypertrace/entity/query/service/client/EntityCacheKey.java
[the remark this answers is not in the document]
tim-mwangi: Ok. Is it a good idea to use the same requestContext that came into server to make client requests? Maybe we would have to create a new one before doing the invocation? I think for graphql it's ok since they don't use the clients defined by the services. They access the API directly.

entity-service-client/src/main/java/org/hypertrace/entity/query/service/client/EntityLabelsClient.java
aaron-steinfeld: > No. They are based on attribute service. So they correspond to the attribute metadata ids. It's expected that entity-service respects that but we do pass them into the query as strings.

If I understood correctly, we may want to change the naming to something like idAttributeKey or something (or ID - depending on if scope is included in the value). Ideally, we should be able to omit these and have entity service look that information up internally though. I guess that requires some widespread changes to the API since it's currently exposed like query service, as a low level sql-style api.

> Performance considerations; avoiding queries into 2 data sources like we were doing before since the labels will be used to annotate the entities - Service and API - wherever they appear on the UI.

Agree, that makes sense - although I think a general entity cache would serve that role just as well and slot in easier.

> Tied to the bullet above, since we will be associating entity labels as foreign columns in other assets/scopes, and we are not writing these to Pinot, then we need a performant way to append labels to results of these other scopes.

This one scares me - is the foreign scope need immediate? I would want to think this one through because we're basically implementing a specialized join for this one attribute. Everywhere else in the platform, we do that at write time - knowing that we want to move it to read time through something like presto eventually. I would agree that doing this generically is out of scope, but I thought we had said it was acceptable and we didn't need that functionality right now? If it is required, I would push to implement it in the same manner as every other join rather than introduce a whole new concept that doesn't solve the problem for other joins.

entity-service-client/src/main/java/org/hypertrace/entity/query/service/client/EntityCacheKey.java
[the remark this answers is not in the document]
aaron-steinfeld: >  Is it a good idea to use the same requestContext that came into server to make client requests?

It should be fine (and less error prone) - the propagation filtering happens on the server call interceptor, so its already good to propagate before we even read it in a server implementation. Is there some other issue you were thinking of?

> I think for graphql it's ok since they don't use the clients defined by the services. They access the API directly.

True - although it's doing it that way because propagating context in this way makes using the generated clients just as easy as using the hand-coded clients. Context could always be propagated, dehydrated and then use the old clients in the same way as they've been used.

Does that make sense? This stuff is a bit confusing so I hope I'm not talking in circles here.

entity-service-client/src/main/java/org/hypertrace/entity/query/service/client/EntityLabelsClient.java
tim-mwangi: > If I understood correctly, we may want to change the naming to something like idAttributeKey or something (or ID - depending on if scope is included in the value). Ideally, we should be able to omit these and have entity service look that information up internally though. I guess that requires some widespread changes to the API since it's currently exposed like query service, as a low level sql-style api.

Actually I will backtrack, sort of. Since entity-service has no knowledge of attributes in attribute service, then these are the attribute names in the doc store config defined for entities in mongo

```
entity.service.attributeMap = [
  {
    "scope": "API",
    "name": "API.id",
    "subDocPath": "entityId"
  },
  {
    "scope": "API",
    "name": "API.name",
    "subDocPath": "entityName"
  },
  {
    "scope": "API",
    "name": "API.apiType",
    "subDocPath": "attributes.API_TYPE"
  },
...
]
``` 
> > Tied to the bullet above, since we will be associating entity labels as foreign columns in other assets/scopes, and we are not writing these to Pinot, then we need a performant way to append labels to results of these other scopes.
> 
> This one scares me - is the foreign scope need immediate? I would want to think this one through because we're basically implementing a specialized join for this one attribute. Everywhere else in the platform, we do that at write time - knowing that we want to move it to read time through something like presto eventually. I would agree that doing this generically is out of scope, but I thought we had said it was acceptable and we didn't need that functionality right now? If it is required, I would push to implement it in the same manner as every other join rather than introduce a whole new concept that doesn't solve the problem for other joins.

The foreign scope need is to be able to do queries on `API_TRACE.apiLabels` for example where labels are not written to Pinot.

We had issues with write time joins since unlike other attributes this one, being an entity attribute, not an attribute on the events associated with the entity can mutate and Pinot does not allow for mutations. Also, the logic to do the joins is the same whether the foreign scope equals primary scope or is different. i.e querying for `API.labels` and `API_TRACE.apiLabels` will go through the same path.

Also, unless I'm confused I thought we agreed to do it this way and avoid writing the labels to Pinot? Writing to Pinot would not be ideal when querying for entities. Consider the services list query for example. When querying from Pinot we group by service id and service name in order to be able to get the aggregated metrics for all the services together with the name and id in one query. If labels were to be involved, since they are an attribute, the results would not be what we want if we grouped by that column - since it's a string array. So we fallback to querying EDS for labels to avoid this weird group by and if doing it for the primary scopes and foreign scopes is the same logic then I'd rather do it this way.
CC @buchi-busireddy @sanjay-nagaraj

aaron-steinfeld: > The foreign scope need is to be able to do queries on API_TRACE.apiLabels for example where labels are not written to Pinot.

I didn't realize querying traces by label was a requirement for the first cut - My impression, perhaps incorrect, was that we first just wanted to be able to see tags on entities and update them.

> We had issues with write time joins since unlike other attributes this one, being an entity attribute, not an attribute on the events associated with the entity can mutate and Pinot does not allow for mutations. 

Didn't we need to do this with isExternal for something around domains? I thought I recalled this being done before. Anyway, I agree with the points about the implications of writing to pinot - I thought we were avoiding that too by not supporting those types of functionality right now, rather than implementing the join ourselves.

aaron-steinfeld: Anyway, since it's already implemented the discussion here was more about limiting the scope so is not super important. Will take a look at what's going on in gateway and how that's used - as long as there's not lots of custom code around labels, we're good.

tim-mwangi: > > The foreign scope need is to be able to do queries on API_TRACE.apiLabels for example where labels are not written to Pinot.
> 
> I didn't realize querying traces by label was a requirement for the first cut - My impression, perhaps incorrect, was that we first just wanted to be able to see tags on entities and update them.

Oh you can do that right now once @arjunlalb puts in the UI changes. Though as it is, if you want to avoid queries to both QS and EDS for a single entity query in gw service, you have to structure your queries in the UI so the labels one is not part of one requesting for metrics. This works on API and Service overview pages.

If you want the labels displayed on the entity list pages, then it gets interesting. I guess you can still make 2 requests but doing the joining at the UI gets tricky. So that's where a cache in gateway service comes in.

> > We had issues with write time joins since unlike other attributes this one, being an entity attribute, not an attribute on the events associated with the entity can mutate and Pinot does not allow for mutations.
> 
> Didn't we need to do this with isExternal for something around domains? I thought I recalled this being done before. Anyway, I agree with the points about the implications of writing to pinot - I thought we were avoiding that too by not supporting those types of functionality right now, rather than implementing the join ourselves.

`isExternal` and `apiDiscoveryState are not big problems since they are simple strings. With labels it's tricky since it's a string array. Also it can mutate.

aaron-steinfeld: > Oh you can do that right now once @arjunlalb puts in the UI changes. Though as it is, if you want to avoid queries to both QS and EDS for a single entity query in gw service, you have to structure your queries in the UI so the labels one is not part of one requesting for metrics. This works on API and Service overview pages.

As far as I know, the only places we're querying labels are on entities queries - either for the entity list or overviews. The UI, at this point (@arjunlalb correct me if I'm wrong) isn't working with labels at the trace level, and the new attributes you added in helm-values are only on entities too.

As to the query structure, is there any benefit of sending two queries on the UI side vs sending on query and making the gateway fire two queries?

I do get the benefit of the cache, it just seems like it should support entities more broadly vs a specific cache for labels - because that implies gateway service requires specific logic for labels too.

entity-service-client/src/main/java/org/hypertrace/entity/query/service/client/EntityLabelsClient.java
```diff
@@ -0,0 +1,34 @@
...
+
+public interface EntityLabelsClient {
+  List<String> getEntityLabelsForEntity(String idColumnName,
+                                        String labelsColumnName,
+                                        String id,
+                                        String type,
+                                        Map<String, String> headers,
+                                        String tenantId);
+
+  void refreshEntityLabelsForEntity(String idColumnName,
```
aaron-steinfeld, line 14: When would refresh be called explicitly?

entity-service-client/src/main/java/org/hypertrace/entity/query/service/client/EntityLabelsCachingClient.java
```diff
@@ -0,0 +1,216 @@
...
+        List<String> labels = row.getColumn(1).getStringArrayList();
+
+        entityLabelsByEntityIdMap.put(entityIdToCacheKey.get(entityId), labels);
+      }
+    }
+
+    return entityLabelsByEntityIdMap;
+  }
+
+  private static Expression.Builder createColumnExpression(String columnName) {
```
aaron-steinfeld, line 192: nit: should return an expression or be renamed

entity-service-client/src/main/java/org/hypertrace/entity/query/service/client/EntityLabelsCachingClient.java
```diff
@@ -0,0 +1,216 @@
...
+    if (!iter.hasNext()) {
+      return Map.of();
+    }
+
+    List<EntityCacheKey<EntityTypeAndId>> entityIdKeys = new ArrayList<>();
+    while (iter.hasNext()) {
+      entityIdKeys.add(iter.next());
+    }
+
+    // The assumption we make is that the all the entity id keys are of the same entity type and
```
aaron-steinfeld, line 136: If we took those in as params, we could unwind that assumption. But really the only way to completely avoid it would be to store it as a tree, where the top level key is the context and then below that we store entities. Otherwise, loadAll could always be invoked with multiple different contexts.

entity-service-client/src/main/java/org/hypertrace/entity/query/service/client/EntityLabelsCachingClient.java
```diff
@@ -0,0 +1,216 @@
...
+    String entityLabelsColumnName = firstEntityCacheKey.getDataKey().getLabelsColumnName();
+    Map<String, String> headers = firstEntityCacheKey.getHeaders();
+
+    Map<String, EntityCacheKey<EntityTypeAndId>> entityIdToCacheKey = new HashMap<>();
+    List<String> entityIds = entityIdKeys.stream()
+        .map(key -> {
+          entityIdToCacheKey.put(key.getDataKey().getId(), key);
+          return key.getDataKey().getId();
+        })
+        .collect(Collectors.toList());
```
aaron-steinfeld, line 150: nit: could we collect this to a map, then get the entityIds just be map.keys()?

entity-service-client/src/main/java/org/hypertrace/entity/query/service/client/EntityLabelsCachingClient.java
```diff
@@ -0,0 +1,216 @@
...
+        .build();
+
+    Iterator<ResultSetChunk> resultSetChunkIterator =
+        entityQueryServiceClient.execute(entityQueryRequest, headers);
+
+    Map<EntityCacheKey<EntityTypeAndId>, List<String>> entityLabelsByEntityIdMap = new HashMap<>();
+    while (resultSetChunkIterator.hasNext()) {
+      ResultSetChunk chunk = resultSetChunkIterator.next();
+
+      if (chunk.getRowCount() < 1) {
```
aaron-steinfeld, line 177: Is this needed? It avoids calling the for loop on an empty list, but doesn't change the logic otherwise.

entity-service-client/src/main/java/org/hypertrace/entity/query/service/client/EntityLabelsCachingClient.java
```diff
@@ -0,0 +1,216 @@
...
+                )
+        )
+        .setLimit(entityIds.size())
+        .build();
+
+    Iterator<ResultSetChunk> resultSetChunkIterator =
+        entityQueryServiceClient.execute(entityQueryRequest, headers);
+
+    Map<EntityCacheKey<EntityTypeAndId>, List<String>> entityLabelsByEntityIdMap = new HashMap<>();
+    while (resultSetChunkIterator.hasNext()) {
```
aaron-steinfeld, line 174: Alternatively could stream and flatmap then collect into a map

entity-service-client/src/main/java/org/hypertrace/entity/query/service/client/EntityLabelsCachingClient.java
jcchavezs: Not necessarily needed but makes it more readable as you know you can stop reading here.

tim-mwangi:
Closing this one out.

PR #29, opened 2020-10-12
fix: include customerId field as well when filtering for tenantId

tim-mwangi:
For old entities - to avoid entityId regeneration and backwards compatibility break - that are never upserted with the newer tenantId

### DIFF
--- a/entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
+++ b/entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
@@ -79,9 +79,26 @@ public class DocStoreConverter {
     return docStoreQuery;
   }
 
+  /**
+   * Since we decided to keep the customerId field around for old entities - to avoid entityId
+   * regeneration and backwards compatibility break - these entities are never
+   * upserted and hence the tenantId column has not been written into the document. So when creating
+   * the filter for tenantId, we need to match on both field names.
+   * @param tenantId
+   * @return OR filter on customerId and tenantId field names
+   */
   public static Filter getTenantIdEqFilter(String tenantId) {
-    return new Filter(Filter.Op.EQ, EntityServiceConstants.TENANT_ID, tenantId);
+    Filter filter = new Filter();
+    filter.setOp(Op.OR);
+    filter.setChildFilters(
+        new Filter[] {
+            new Filter(Filter.Op.EQ, EntityServiceConstants.TENANT_ID, tenantId),
+            new Filter(Filter.Op.EQ, EntityServiceConstants.CUSTOMER_ID, tenantId)
+        });
+
+    return filter;
   }
+
 
   private static Filter transform(AttributeFilter filter) {
     try {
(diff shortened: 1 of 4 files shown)


entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
```diff
@@ -79,10 +79,27 @@
...
+   * @return OR filter on customerId and tenantId field names
+   */
   public static Filter getTenantIdEqFilter(String tenantId) {
-    return new Filter(Filter.Op.EQ, EntityServiceConstants.TENANT_ID, tenantId);
+    Filter filter = new Filter();
+    filter.setOp(Op.OR);
+    filter.setChildFilters(
+        new Filter[] {
+            new Filter(Filter.Op.EQ, EntityServiceConstants.TENANT_ID, tenantId),
+            new Filter(Filter.Op.EQ, EntityServiceConstants.CUSTOMER_ID, tenantId)
```
kotharironak, line 96: Changes look fine to me in the context of supporting old entities. But, I think, 

1. Can this be handle as one time DB task by moving old entities with replacing <customer_id> -> <tenanat_id>?
2. What if we purged the old entities?

avinashkolluru: Agree with Ronak. Why not a one time migration instead of living with this code for ever.

tim-mwangi: Sure. Have to do it for all environments upto prod. Also I don't know if you read my comment about entityId generation depending on "customerId" field itself. So if you do this you make all current data in Pinot and deep store obsolete since the entityId generation depends on that field.

avinashkolluru: May be just a one time rename of customerId to tenantId? 
https://docs.mongodb.com/manual/reference/operator/update/rename/

sanjay-nagaraj: How about we remove this code after a few weeks.

avinashkolluru: @sanjay-nagaraj The problem is not with the entityId that flows into pinot. The issue here is that some old mongo docs have a field named 'customerId' and the new ones have 'tenantId'

This OR condition is being applied to support the filtering on the customerId/tenantId field for older docs.

buchi-busireddy: We had this exact change when we migrated from `customerId` to `tenantId` to provide backward compatibility and I've removed the change once I thought all the active entities have `tenantId` populated in them in all environments. The migration to `tenantId` was done a while ago so I wouldn't expect any latest entities to not have it, unless someone used old code and created entities.
@tim-mwangi let's take this offline and we should understand why the entities don't have the `tenantId` field. I'll work with you on that.

buchi-busireddy: FTR, we're fixing this manually in all envs with Mongo's rename command since it's one-off case and isn't worth spending lot of time on. Thanks @tim-mwangi for reporting and the changes.

tim-mwangi:
Closing this one since @buchi-busireddy renamed the field to tenantId in the old entities.

PR #32, opened 2020-10-16
Adding a new version of EntityTypeService, which defines entity type based on attributes defined in AttributeService

buchi-busireddy:
instead of redefining the full entity on its own.

This will help us to close the loop on making the Attributes as the core
metadata across the entire Hypertrace platform and every service refering
to them.


entity-service-api/src/main/proto/org/hypertrace/entity/type/service/v2/entity_type_service.proto
```diff
@@ -0,0 +1,22 @@
+syntax = "proto3";
+
+option java_multiple_files = true;
+package org.hypertrace.entity.type.service.v2;
+
+message EntityType {
+  string name = 1;
```
aaron-steinfeld, line 7: what's the difference between the name and scope? I guess it doesn't hurt to separate them, but we'd expect them to be the same right?

entity-service-api/src/main/proto/org/hypertrace/entity/type/service/v2/entity_type_service.proto
```diff
@@ -0,0 +1,22 @@
...
+}
+
+message EntityTypeFilter {
+  repeated string name = 1;
+}
+
+message Empty {}
+
+service EntityTypeServiceV2 {
+  rpc upsertEntityType (EntityType) returns (EntityType) {}
```
aaron-steinfeld, line 19: let's make sure these pass the buf conventions as new apis - I don't believe they do.

entity-service-client/src/main/java/org/hypertrace/entity/type/service/v2/client/EntityTypeServiceClient.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.List;
+import java.util.concurrent.Callable;
+import org.hypertrace.core.grpcutils.client.GrpcClientRequestContextUtil;
+import org.hypertrace.core.grpcutils.client.RequestContextClientCallCredsProviderFactory;
+import org.hypertrace.entity.type.service.v2.EntityType;
+import org.hypertrace.entity.type.service.v2.EntityTypeFilter;
+import org.hypertrace.entity.type.service.v2.EntityTypeServiceV2Grpc;
+import org.hypertrace.entity.type.service.v2.EntityTypeServiceV2Grpc.EntityTypeServiceV2BlockingStub;
+
+public class EntityTypeServiceClient {
```
aaron-steinfeld, line 14: we can probably drop the client - this one doesn't provide any value beyond the grpc generated ones.

entity-service-impl/src/main/java/org/hypertrace/entity/type/service/v2/model/EntityTypeDocument.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.Objects;
+import org.hypertrace.core.documentstore.Document;
+import org.hypertrace.entity.service.constants.EntityServiceConstants;
+import org.hypertrace.entity.type.service.v2.EntityType;
+
+public class EntityTypeDocument implements Document {
+  // Since there could be additional metadata fields written by the doc store,
+  // ignore them while reading the object from JSON.
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper().configure(
+      DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);;
```
aaron-steinfeld, line 16:
```suggestion
      DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```

entity-service-impl/src/main/java/org/hypertrace/entity/type/service/v2/model/EntityTypeDocument.java
```diff
@@ -0,0 +1,117 @@
...
+  @JsonProperty
+  private String attributeScope;
+
+  @JsonProperty
+  private String idAttributeKey;
+
+  public EntityTypeDocument() {}
+
+  public EntityTypeDocument(String name, String attributeScope, String idAttributeKey) {
+    this(null, name, attributeScope, idAttributeKey);
```
aaron-steinfeld, line 33: null tenantId? that seems bad - should we require the tenantId in the fromProto call (or a request context or something?)

entity-service-api/src/main/proto/org/hypertrace/entity/type/service/v2/entity_type_service.proto
```diff
@@ -0,0 +1,22 @@
...
+  rpc deleteEntityTypes (EntityTypeFilter) returns (Empty) {}
+  rpc queryEntityTypes (EntityTypeFilter) returns (stream EntityType) {}
```
aaron-steinfeld, line 21: there's probably no real benefit to streaming the result here.

entity-service-impl/src/main/java/org/hypertrace/entity/type/service/v2/EntityTypeServiceImplV2.java
```diff
@@ -0,0 +1,122 @@
...
+    if (tenantId.isEmpty()) {
+      responseObserver.onError(new ServiceException("Tenant id is missing in the request."));
+      return;
+    }
+
+    try {
+      // Set tenant id.
+      EntityTypeDocument document = EntityTypeDocument.fromProto(request);
+      document.setTenantId(tenantId.get());
+      entityTypeCollection.upsert(new SingleValueKey(tenantId.get(), request.getName()), document);
```
aaron-steinfeld, line 49: I would expect upsert to return the upserted value rather than the requested value - what's the behavior if this entity type already exists with different values? Or a superset of values?

entity-service-api/src/main/proto/org/hypertrace/entity/type/service/v2/entity_type_service.proto
```diff
@@ -0,0 +1,22 @@
...
+  string id_attribute_key = 3;
+}
+
+message EntityTypeFilter {
+  repeated string name = 1;
+}
+
+message Empty {}
+
+service EntityTypeServiceV2 {
```
aaron-steinfeld, line 18: Can we remove the v2 from the name? Even if it's annoying to fully qualify it in one spot now, it allows us to fully cut over later without another breaking change.

entity-service-api/src/main/proto/org/hypertrace/entity/type/service/v2/entity_type_service.proto
```diff
@@ -0,0 +1,22 @@
+syntax = "proto3";
+
+option java_multiple_files = true;
+package org.hypertrace.entity.type.service.v2;
+
+message EntityType {
+  string name = 1;
+  string attribute_scope = 2;
+  string id_attribute_key = 3;
```
aaron-steinfeld, line 9: I think we need name_attribute_key as well, given the current entity design which has a first class name field.

entity-service-api/src/main/proto/org/hypertrace/entity/type/service/v2/entity_type_service.proto
buchi-busireddy: Yes, same for the cases we have so far. When we bring in infra entities, it might change though hence keeping it.

entity-service-api/src/main/proto/org/hypertrace/entity/type/service/v2/entity_type_service.proto
buchi-busireddy: Done.

entity-service-client/src/main/java/org/hypertrace/entity/type/service/v2/client/EntityTypeServiceClient.java
buchi-busireddy: The way I see, these are convenience methods so that the clients don't have to create requests, etc. I find it pretty useful.

entity-service-impl/src/main/java/org/hypertrace/entity/type/service/v2/model/EntityTypeDocument.java
buchi-busireddy: Exactly thought the same. Fixed.

entity-service-api/src/main/proto/org/hypertrace/entity/type/service/v2/entity_type_service.proto
buchi-busireddy: Fixed.

entity-service-impl/src/main/java/org/hypertrace/entity/type/service/v2/EntityTypeServiceImplV2.java
buchi-busireddy: Done

entity-service-api/src/main/proto/org/hypertrace/entity/type/service/v2/entity_type_service.proto
buchi-busireddy: Done

entity-service-api/src/main/proto/org/hypertrace/entity/type/service/v2/entity_type_service.proto
buchi-busireddy: Done

entity-service-client/src/main/java/org/hypertrace/entity/type/service/v2/client/EntityTypeServiceClient.java
aaron-steinfeld: It doesn't hurt anything to leave it, but my concern with it is that it pushes a few different decision on the user:
- It makes the context explicit. If I have to pass in tenantId explicitly, then changing the context we propagate through the system becomes more difficult. It also means as a caller I have to unwrap the context, just to pass it to the client to repackage.
- It makes all of the calls using a blocking stub. One of the nice things about grpc is that it generates feature, streaming and blocking stubs right out of the box. Leave it to the caller to decide.
- It prevents using grpc features like deadlines, executors etc.
- It exposes a larger API surface than we've declared in the proto file, by way of convenience methods (e.g. deleteAll - if that should be part of the api, let's declare it in the proto file so we can do breaking change detection against it)

aaron-steinfeld (dismissed):

entity-service/src/integrationTest/java/org/hypertrace/entity/service/service/EntityTypeServiceV2Test.java
```diff
@@ -6,17 +6,17 @@
...
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 /**
- * Integration test for testing {@link org.hypertrace.entity.type.service.v2.EntityTypeServiceImplV2}
+ * Integration test for testing {@link EntityTypeServiceImpl}
  */
 public class EntityTypeServiceV2Test {
```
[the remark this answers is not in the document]
buchi-busireddy: Done

entity-service-client/src/main/java/org/hypertrace/entity/type/service/v2/client/EntityTypeServiceClient.java
buchi-busireddy: :) Done

aaron-steinfeld (approved):

entity-service-impl/src/main/java/org/hypertrace/entity/type/service/v2/model/EntityTypeDocument.java
```diff
@@ -0,0 +1,127 @@
...
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    EntityTypeDocument document = (EntityTypeDocument) o;
```
tim-mwangi, line 106: Much of this has not been unit tested

entity-service-impl/src/main/java/org/hypertrace/entity/type/service/v2/EntityTypeServiceImpl.java
```diff
@@ -0,0 +1,151 @@
...
+        // Return an error for now.
+        responseObserver.onError(new RuntimeException("Error upserting EntityType:" + request));
+      }
+    } catch (IOException e) {
+      responseObserver.onError(new RuntimeException("Error upserting EntityType:" + request, e));
+    }
+  }
+
+  private boolean isValidUpsert(EntityType entityType) {
+    return !Strings.isNullOrEmpty(entityType.getName()) &&
```
tim-mwangi, line 77: Much of this has not been tested. If it is easy to test can we put in the tests?

tim-mwangi (approved):

entity-service-impl/src/main/java/org/hypertrace/entity/type/service/v2/EntityTypeServiceImpl.java
buchi-busireddy: FTR, discussed offline. Follow up PRs will address the unit tests. This PR already covers all the code with integration tests.

PR #33, opened 2020-10-16
fix: filtering based on entity labels

tim-mwangi:
`labels EQUALS "foo"` translates to labels contains "foo" and AND chain means that the labels should contain all of the labels on the RHS side.
`labels IN ["foo","bar"]` translates to labels contains at least one of the labels in the RHS side of the filter.

### DIFF
--- a/entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
+++ b/entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
@@ -89,29 +89,25 @@ public class DocStoreConverter {
   }
 
   private static Filter transform(AttributeFilter filter) {
-    try {
-      if (filter.hasAttributeValue()) {
-        if (ATTRIBUTES_LABELS_FIELD_NAME.equals(filter.getName()) && filter.getOperator() == Operator.EQ) {
-          return transformToEqFilterWithValueListRhs(filter);
-        } else if (ATTRIBUTES_LABELS_FIELD_NAME.equals(filter.getName()) && filter.getOperator() == Operator.IN) {
-          return transformToOrFilterChainForStrArray(filter);
-        } else {
-          return transformNonListRhsFilterTypes(filter);
-        }
-      } else { // AND or OR filter chains
-        Filter f = new Filter();
-        f.setFieldName(filter.getName());
-        f.setOp(transform(filter.getOperator()));
-
-        f.setChildFilters(
-            filter.getChildFilterList().stream()
-                .map(DocStoreConverter::transform)
-                .collect(Collectors.toList())
-                .toArray(new Filter[]{}));
-        return f;
+    if (filter.hasAttributeValue()) {
+      if (ATTRIBUTES_LABELS_FIELD_NAME.equals(filter.getName()) && filter.getOperator() == Operator.EQ) {
+        return transformToEqFilterWithValueListRhs(filter);
+      } else if (ATTRIBUTES_LABELS_FIELD_NAME.equals(filter.getName()) && filter.getOperator() == Operator.IN) {
+        return transformToOrFilterChainForStrArray(filter);
+      } else {
+        return transformNonListRhsFilterTypes(filter);
       }
-    } catch (IOException ioe) {
-      throw new IllegalArgumentException(String.format("transform: Error converting filter for query. Filter:  %s", filter), ioe);
+    } else { // AND or OR filter chains
+      Filter f = new Filter();
+      f.setFieldName(filter.getName());
+      f.setOp(transform(filter.getOperator()));
+
+      f.setChildFilters(
+          filter.getChildFilterList().stream()
+              .map(DocStoreConverter::transform)
+              .collect(Collectors.toList())
+              .toArray(new Filter[]{}));
+      return f;
     }
   }
 
@@ -143,22 +139,7 @@ public class DocStoreConverter {
 
     f.setChildFilters(
         attributeFilter.getAttributeValue().getValueList().getValuesList().stream()
-            .map(rhsAttributeValue -> {
-              Filter childFilter = new Filter();
-              childFilter.setFieldName(fieldName);
-              childFilter.setOp(Op.EQ);
-
-              JsonNode mapNode = null;
-              try {
-                mapNode = OBJECT_MAPPER
-                    .readTree(JSONFORMAT_PRINTER.print(rhsAttributeValue));
-              } catch (JsonProcessingException | InvalidProtocolBufferException e) {
-                throw new RuntimeException(e);
-              }
-              Map map = OBJECT_MAPPER.convertValue(mapNode, Map.class);
-              childFilter.setValue(map);
-              return childFilter;
-            })
+            .map(rhsAttributeValue -> createEqFilterForAttributeValue(fieldName, rhsAttributeValue))
             .collect(Collectors.toList())
             .toArray(new Filter[]{})
     );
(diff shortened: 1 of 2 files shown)


entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
```diff
@@ -89,8 +93,18 @@ private static Filter transform(AttributeFilter filter) {
       f.setFieldName(filter.getName());
       f.setOp(transform(filter.getOperator()));
       if (filter.hasAttributeValue()) {
-        transform(filter.getAttributeValue(), f, isPartOfAttributeMap(filter.getName()));
+        //transform(filter.getAttributeValue(), f);
+        if (ATTRIBUTES_LABELS_FIELD_NAME.equals(f.getFieldName()) && f.getOp() == Op.EQ) {
```
aaron-steinfeld, line 97: So this would eventually go away right? Once entity service is attribute-aware and can map the LHS to an attribute type? If yes, might be worth moving it into its own function rather than inlining it.

aaron-steinfeld: How does this logic interface with the logic on line 193 for a value list? Is that logic based on the RHS rather than LHS?

buchi-busireddy, line 97: So, because we don't have type concept in ES today, we are hard-coding this for only labels attribute right?

entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
```diff
@@ -106,6 +120,61 @@ private static boolean isPartOfAttributeMap(String fieldName) {
...
+
+              JsonNode mapNode = null;
+              try {
+                mapNode = OBJECT_MAPPER
+                    .readTree(JSONFORMAT_PRINTER.print(rhsAttributeValue));
+              } catch (JsonProcessingException | InvalidProtocolBufferException e) {
+                throw new RuntimeException(e);
+              }
+              Map map = OBJECT_MAPPER.convertValue(mapNode, Map.class);
+              f.setValue(map);
```
buchi-busireddy, line 153: I'm not sure if I understand this fully. Can you give an example please?

entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
```diff
@@ -89,8 +93,18 @@ private static Filter transform(AttributeFilter filter) {
       f.setFieldName(filter.getName());
       f.setOp(transform(filter.getOperator()));
       if (filter.hasAttributeValue()) {
-        transform(filter.getAttributeValue(), f, isPartOfAttributeMap(filter.getName()));
+        //transform(filter.getAttributeValue(), f);
+        if (ATTRIBUTES_LABELS_FIELD_NAME.equals(f.getFieldName()) && f.getOp() == Op.EQ) {
+          transformToEqFilterWithValueListRhs(filter.getAttributeValue(), f);
+        } else if (ATTRIBUTES_LABELS_FIELD_NAME.equals(f.getFieldName()) && f.getOp() == Op.IN) {
+          transformToOrFilterChainForStrArray(filter.getAttributeValue(), f);
+          // Early return on this one since the childFilters are in the Or chain.
```
aaron-steinfeld, line 101: Why does this return early but none of the others? I'm guessing from the comment because this transformation involves childfilters we don't want to overwrite them later. would it be feasible the logic so it doesn't behave differently? Would there ever even be childfilters if line 95 were true or could the child filter logic move into an else clause?

entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
```diff
@@ -106,6 +120,61 @@ private static boolean isPartOfAttributeMap(String fieldName) {
...
+//      transformToEqFilterWithValueListRhs(attributeValue, filter);
+//    } else if (ATTRIBUTES_LABELS_FIELD_NAME.equals(filter.getFieldName()) && filter.getOp() == Op.IN) {
+//      transformToOrFilterChainForStrArray(attributeValue, filter);
+//    } else {
+//      transform(attributeValue, filter, isPartOfAttributeMap(filter.getFieldName()));
+//    }
+//  }
+
+  private static void transformToOrFilterChainForStrArray(AttributeValue attributeValue, Filter filter) {
+    String fieldName = filter.getFieldName() + "." + "valueList" + "." + "values";
```
aaron-steinfeld, line 134: why are we concatenating literal strings?

entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
tim-mwangi: yes

entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
tim-mwangi: What you want to have on the rhs is the same object format that's written for AttributeValue for label. So this:
`"attributes.labels.valueList.values": { "value": { "string": "foo" } }`. The JSON written down into mongo for the labels is `{'"valueList": {"values": [{ "value": { "string": "foo" } }, { "value": { "string": "bar" } }]}}`

entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
tim-mwangi: > So this would eventually go away right? Once entity service is attribute-aware and can map the LHS to an attribute type? If yes, might be worth moving it into its own function rather than inlining it.

Yes. Will go away. Can be refactored to be cleaner.

> How does this logic interface with the logic on line 193 for a value list? Is that logic based on the RHS rather than LHS?
 It doesn't. Since we need the full JSON object and not just the string or whichever primitive value is wrapped in the proto, we need different logic to create the RHS for labels.

entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
aaron-steinfeld: I guess I was more asking why `"." + "valueList" + "." + "values"` vs `".valueList.values"`

tim-mwangi: lol. copy paste?

tim-mwangi: fixed.

entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
```diff
@@ -84,28 +89,108 @@ public static Filter getTenantIdEqFilter(String tenantId) {
   }
 
   private static Filter transform(AttributeFilter filter) {
+    try {
+      if (filter.hasAttributeValue()) {
+        if (ATTRIBUTES_LABELS_FIELD_NAME.equals(filter.getName()) && filter.getOperator() == Operator.EQ) {
```
avinashkolluru, line 94: Can this be generalized for every multivalued/list attribute instead of special casing for labels?

tim-mwangi: No since we don't know the type of the field values in entity-service. We need to type the attributes used in entity-service and that involves having the entity-service connect to and make use of attribute service. For now we can get around this by treating attribute.labels specially.

entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
```diff
@@ -84,28 +89,108 @@ public static Filter getTenantIdEqFilter(String tenantId) {
...
+    String fieldName = attributeFilter.getName() + VALUE_LIST_VALUES_CONST;
+
+    Filter f = new Filter();
+    f.setFieldName("");
+    f.setOp(Op.OR);
+
+    f.setChildFilters(
+        attributeFilter.getAttributeValue().getValueList().getValuesList().stream()
+            .map(rhsAttributeValue -> {
+              Filter childFilter = new Filter();
```
buchi-busireddy, line 147: how's this block different from the below `transformToEqFilterWithValueListRhs()`? In that method also you seem to be using only AttributeValue. Can you reuse the code probably?

tim-mwangi: Mmh let me take a look. There is a reason in the back of my head but may be not.

tim-mwangi: No reason actually. Will refactor.

buchi-busireddy (approved):

PR #34, opened 2020-10-28
ci: add dockerhub auth

ravisingal:


tim-mwangi (approved):

.circleci/config.yml
```diff
@@ -65,6 +74,9 @@ jobs:
       - setup_build_environment
       - setup_remote_docker
       - populate_and_save_cache
+      - run:
```
aaron-steinfeld, line 77: Not necessary, but it'd be nice to move this into its own command rather than duplicating in each job (see: https://github.com/hypertrace/attribute-service/pull/28/files )

aaron-steinfeld (approved):

.circleci/config.yml
ravisingal: done.

aaron-steinfeld (approved):

tim-mwangi (approved):

entity-service-api/build.gradle.kts
```diff
@@ -59,4 +59,8 @@ dependencies {
   api("io.grpc:grpc-protobuf:1.32.1")
   api("io.grpc:grpc-stub:1.32.1")
   api("javax.annotation:javax.annotation-api:1.3.2")
+
```
aaron-steinfeld, line 62: Guava is not a runtime only dep, this constraint should match how the dependency is included. But in general, we can refer snyk failures to a separate PR, no need to handle them with the dockerhub changes

ravisingal: `snyk-scan` is a required test and it won't allow merge without fixing it.

aaron-steinfeld: we can override that in this case, let's not mix changes. The guava change here has not insignificant implications and a decent chance of causing runtime issues (both because we're changing the version at runtime so it differs from compile tie, and because we're using the android version which is missing java 8 support)

ravisingal: reverted the snyk related changes. will revert in other repos as well.

ravisingal:
@aaron-steinfeld please remove `snyk-scan` from required check.

aaron-steinfeld (approved):
Done

tim-mwangi (approved):

PR #41, opened 2020-11-10
refactor: add support for upsert of v1 or v2 typed entities

aaron-steinfeld:
The purpose of this change is to allow the upsert API to support entities formed by v1 entity types in addition to the existing v1 support. Both produce the same entity objects, but the data in those entities looks a bit different. V2 typed entities will always have an id, and don't use identifying attributes. V1 typed entities are the reverse, requiring identifying attributes and using those in place of an ID. The move to v2 entity types is to support generating ids dynamically and allowing code to generate and work with many entity attributes, including ID without calling into entity service.

### DIFF
--- a/entity-service-impl/src/test/java/org/hypertrace/entity/data/service/DocumentParserTest.java
+++ b/entity-service-impl/src/test/java/org/hypertrace/entity/data/service/DocumentParserTest.java
@@ -1,0 +1,59 @@
+package org.hypertrace.entity.data.service;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import com.google.protobuf.InvalidProtocolBufferException;
+import com.google.protobuf.util.JsonFormat;
+import java.io.IOException;
+import java.util.Optional;
+import java.util.UUID;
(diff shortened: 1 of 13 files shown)


entity-service-impl/src/test/java/org/hypertrace/entity/data/service/DocumentParserTest.java
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  @Test
+  void returnsMessageIfAbleToParseOrThrow() throws InvalidProtocolBufferException {
+    assertEquals(ENTITY, PARSER.parseOrThrow(GOOD_JSON, Entity.newBuilder()));
+  }
+
+  @Test
+  void returnsMessageOptionalIfAbleToParseOrLog() {
+
```
tim-mwangi, line 51: nit: space

tim-mwangi (approved):

entity-service-impl/src/test/java/org/hypertrace/entity/data/service/DocumentParserTest.java
aaron-steinfeld: I'll take it out in the other entity svc PR when I rebase it.

PR #52, opened 2020-12-02
Adds limit and offset implementation

inespo:
https://github.com/hypertrace/entity-service/issues/57
Need this implemented to allow batch fetching for large entities volume

EntityQueryService has limit and offset specified in the proto as part
of the Query message. However, there's no implementation and it's has been
faux attributes.

This adds the implementation of order, limit and offset

## Description
Please include a summary of the change, motivation and context.

### Testing

### Checklist:

### Documentation

### DIFF
--- a/entity-service-api/src/main/proto/org/hypertrace/entity/data/service/v1/entity_data_request.proto
+++ b/entity-service-api/src/main/proto/org/hypertrace/entity/data/service/v1/entity_data_request.proto
@@ -40,6 +40,8 @@ message Query {
   string entity_type = 4;
   string entity_name = 5;
   AttributeFilter filter = 6;
+  int32 limit = 7;
+  int32 offset = 8;
 }
 
 message RelationshipsQuery {
--- a/entity-service-api/src/main/proto/org/hypertrace/entity/data/service/v1/entity_data_request.proto
+++ b/entity-service-api/src/main/proto/org/hypertrace/entity/data/service/v1/entity_data_request.proto
@@ -40,6 +40,7 @@ message Query {
   string entity_type = 4;
   string entity_name = 5;
   AttributeFilter filter = 6;
+  repeated OrderByExpression order_by = 9;
   int32 limit = 7;
   int32 offset = 8;
 }
@@ -55,4 +56,14 @@ message AttributeFilter {
   Operator operator = 2;
   AttributeValue attributeValue = 3;
   repeated AttributeFilter childFilter = 4;
+}
+
+message OrderByExpression {
+  string name = 1;
+  SortOrder order = 2;
+}
+
+enum SortOrder {
+  ASC = 0;
+  DESC = 1;
 }
--- a/entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/entity_query_request.proto
+++ b/entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/entity_query_request.proto
@@ -12,7 +12,7 @@ message EntityQueryRequest {
   string entityType = 2;
   Filter filter = 3;
   repeated Expression selection = 4;
-
+  repeated OrderByExpression orderBy = 7;
   int32 limit = 5;
   int32 offset = 6;
 }
--- a/entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryConverter.java
+++ b/entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryConverter.java
@@ -32,6 +32,17 @@ public class EntityQueryConverter {
     if (attributeFilter != null) {
       queryBuilder.setFilter(attributeFilter);
     }
+
+    int limit = queryRequest.getLimit();
+    if (limit > 0) {
+      queryBuilder.setLimit(limit);
+    }
+
+    int offset = queryRequest.getOffset();
+    if (offset > 0) {
+      queryBuilder.setOffset(offset);
+    }
+
     return queryBuilder.build();
   }
 
--- a/entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryConverter.java
+++ b/entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryConverter.java
@@ -2,8 +2,11 @@ package org.hypertrace.entity.query.service;
 
 import static java.util.stream.Collectors.toMap;
 
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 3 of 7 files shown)


inespo:
cc @adriancole

jcchavezs:
If I understand correctly, in this context offset allows to retrieve records that are default ordered? In general I don't like offsets because they aren't deterministic nor efficient. They aren't deterministic because you basically ask: give me `x` records skipping `y` first and those `y` skipped records can change from call to call and consequently those `x` can too (even with the same offset). On efficiency, this is a problem when you do filtering and ordering because you first filter, order and then skip so imagine you retrieve 1M records to skip 999K. Finally, offset is usually coupling to the data storage as not every DB supports it (e.g. Elasticsearch where you probably need to lookup the record previous to the offset and use the `from`)

In sum, maybe none of this is an actual problem in this service but it would be good to understand with what we are committing to if we implement offset.

tim-mwangi:
> If I understand correctly, in this context offset allows to retrieve records that are default ordered? In general I don't like offsets because they aren't deterministic nor efficient. They aren't deterministic because you basically ask: give me `x` records skipping `y` first and those `y` skipped records can change from call to call and consequently those `x` can too (even with the same offset). On efficiency, this is a problem when you do filtering and ordering because you first filter, order and then skip so imagine you retrieve 1M records to skip 999K. Finally, offset is usually coupling to the data storage as not every DB supports it (e.g. Elasticsearch where you probably need to lookup the record previous to the offset and use the `from`) so a link to the `next page` (i.e. the records after the last one) is usually desirable.
> 
> In sum, maybe none of this is an actual problem in this service but it would be good to understand with what we are committing to if we implement offset.

Yeah. Limit and offset are interesting and should probably not be there without a concept of ordering which the request should specify. This is to guarantee the same results when the same request is replayed. If the client does not specify ordering then we just return whatever the DB gives back. So the onus is on the client to specify the order.

jcchavezs:
Also I am not 100% sure the "default storage sorting" is a thing on
distributed databases so this is also coupling with the implementation.

inespo:
I agree that ideally limit and offset needs to be provided with OrderBy. I wanted to close the gap of faux attributes with this and add the OrderBy in separate PR since it changes the API itself. 

@jcchavezs , "skipped records can change from call to call" this shouldn't happen if you specify an order. Maybe I'm missing something? and ElasticSearch and it supports pagination with `from` as a numeric value. But, nonetheless, I think that you have a valid point which direction we want to go with supporting pagination.

entity-service-api/src/main/proto/org/hypertrace/entity/data/service/v1/entity_data_request.proto
```diff
@@ -54,3 +57,13 @@ message AttributeFilter {
...
   repeated AttributeFilter childFilter = 4;
 }
+
+message OrderByExpression {
+  string name = 1;
+  SortOrder order = 2;
+}
+
+enum SortOrder {
+  ASC = 0;
```
buchi-busireddy, line 67: Do you want ASC to be the default?

entity-service-api/src/main/proto/org/hypertrace/entity/data/service/v1/entity_data_request.proto
```diff
@@ -54,3 +57,13 @@ message AttributeFilter {
   AttributeValue attributeValue = 3;
   repeated AttributeFilter childFilter = 4;
 }
+
+message OrderByExpression {
```
buchi-busireddy, line 61: Since this is being used in EQS as well, do you want to extract into a separate file?

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryConverter.java
```diff
@@ -32,6 +32,17 @@ public static Query convertToEDSQuery(EntityQueryRequest queryRequest,
     if (attributeFilter != null) {
       queryBuilder.setFilter(attributeFilter);
     }
+
+    int limit = queryRequest.getLimit();
+    if (limit > 0) {
```
buchi-busireddy, line 37: what happens if you set without checking for the condition?

entity-service-api/src/main/proto/org/hypertrace/entity/data/service/v1/entity_data_request.proto
inespo: I prefer to keep the API separate with different object, we don't want to mix them since the use case would very be different. The code repetition worth keeping the 2 API separate

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryConverter.java
inespo: hmm, you're right the check may not be needed since it will be converted to 0 as well in EDS query.

entity-service-api/src/main/proto/org/hypertrace/entity/data/service/v1/entity_data_request.proto
inespo: That's what I'm thinking. 

If it's unspecified we assumes ascending. Otherwise, the API won't be stringent enough. supposed I have: 
SortOrder {
   Unspecified = 0; 
   Asc = 1;
   Desc = 2;
}

If it's unspecified, then, in the implementation it would have to make it as Ascending Or throw a RuntimeException such as IllegalArgumentException. This is difficult to test for the user of this API.

entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/entity_query_request.proto
```diff
@@ -12,7 +12,7 @@ message EntityQueryRequest {
   string entityType = 2;
   Filter filter = 3;
   repeated Expression selection = 4;
-
+  repeated OrderByExpression orderBy = 7;
```
buchi-busireddy, line 15: You're saying you don't want to reuse b/w EDS and EQS but you're reusing it here right?

inespo: It is not, it's using the OrderByExpression definition from query/service/v1/request.proto. 
Definition is : 
`message OrderByExpression {
  Expression expression = 1;
  SortOrder order = 2;
}
`
While in EDS it uses the one in data/service/v1/entity_data_request.proto. The definition is : 
```
message OrderByExpression {
  string name = 1;
  SortOrder order = 2;
}
```

buchi-busireddy (approved):

Issue #60, opened 2020-12-17
Add a new API to return previous entity while doing bulk upsert


buchi-busireddy:
Use case:
----------
Some of the clients need to know whether an entity has changed or not as part of bulk upsert. Since there is no change data capture support currently, there is no way for clients to implement this. By adding API(s) to return the state of entity before the upsert, client can do the checks itself to see whether the entity got changed or not.

buchi-busireddy:
Implemented in #61

PR #61, opened 2020-12-17
#60 Adding a new API which returns the previous entities before bulk upsert

buchi-busireddy:
## Description
See #60 for more details.

### Testing
Integration tests are verifying the full logic.

### Checklist:

This PR depends on https://github.com/hypertrace/document-store/pull/18 so build will fail until that's merged.

### DIFF
--- a/entity-service-api/src/main/proto/org/hypertrace/entity/data/service/v1/entity_data_service.proto
+++ b/entity-service-api/src/main/proto/org/hypertrace/entity/data/service/v1/entity_data_service.proto
@@ -12,6 +12,7 @@ service EntityDataService {
   }
   rpc upsertEntities (Entities) returns (Empty) {
   }
+  rpc getAndUpsertEntities (Entities) returns (stream Entity) {}
   rpc delete (ByIdRequest) returns (Empty) {
   }
   rpc getById (ByIdRequest) returns (Entity) {
--- a/entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java
+++ b/entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java
@@ -148,7 +148,7 @@ public class EntityDataServiceImpl extends EntityDataServiceImplBase {
         documentMap.put(key, doc);
       }
 
-      Streams.stream(entitiesCollection.returnAndBulkUpsert(documentMap))
+      Streams.stream(entitiesCollection.bulkUpsertAndReturnOlderDocuments(documentMap))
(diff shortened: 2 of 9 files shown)


avinashkolluru (approved):

entity-service-api/src/main/proto/org/hypertrace/entity/data/service/v1/entity_data_service.proto
```diff
@@ -12,6 +12,7 @@ service EntityDataService {
   }
   rpc upsertEntities (Entities) returns (Empty) {
   }
+  rpc getAndUpsertEntities (Entities) returns (stream Entity) {}
```
kotharironak, line 15:
```suggestion
  rpc getAndUpsertEntities (Entities) returns (stream Entity) {
  }
```

kotharironak: As rest of the methods are following that pattern.

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java
```diff
@@ -127,6 +127,41 @@ public void upsertEntities(Entities request, StreamObserver<Empty> responseObser
...
+        SingleValueKey key = new SingleValueKey(tenantId, entry.getKey());
+        documentMap.put(key, doc);
+      }
+
+      Streams.stream(entitiesCollection.bulkUpsertAndReturnOlderDocuments(documentMap))
+          .flatMap(document -> PARSER.<Entity>parseOrLog(document, Entity.newBuilder()).stream())
+          .map(Entity::toBuilder)
+          .map(builder -> builder.setTenantId(tenantId))
+          .map(Entity.Builder::build)
+          .forEach(responseObserver::onNext);
```
kotharironak, line 156: This will send one document at a time, right? Should we send them in batch?

kotharironak (approved):
lgtm. Had minor comments.

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java
buchi-busireddy: Streaming one at a time will allow the client to process it and discard instead of keeping it in memory. i think that's the pattern we should always follow.

entity-service-api/src/main/proto/org/hypertrace/entity/data/service/v1/entity_data_service.proto
buchi-busireddy: I think this is very minor. Will avoid one more commit and full build cycle for this :)

Issue #58, opened 2020-12-14
Supports an operation to check if an attribute exists or not


inespo:
Entity Service has an attribute field that's a map. However, there's no support to check if an attribute has been inserted or not into the map.

Document Store already support Exists operator, and it could be the solution it just exposes the operator and pass it down to Document store.

kotharironak:
The use case here is to fetch entities on certain missing attributes.  As an example, 
`Select all APIs where state = xyz and dosn't have attribute.x`

PR #62, opened 2020-12-17
adds supports for exists/not-exists filters

kotharironak:
## Description
Addressed the issue: https://github.com/hypertrace/entity-service/issues/58

Add supports for checking if the corresponding field exists in the entity.
- exists: returns matching entities documents that contain a request field
- not_exists: returns list of matching entities documents that don't contain requested field

Note:
The underlying document store already supports the exists/non-exists filter:
- https://github.com/hypertrace/document-store/blob/main/document-store/src/main/java/org/hypertrace/core/documentstore/Filter.java
- Mongo implementation provides its impl - https://github.com/hypertrace/document-store/blob/main/document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
- For Postgres, there are few operators not implemented. will create an issue for that.

### Testing
Adds a corresponding integration test.

### Checklist:

### Documentation
Not requires explicitly, can be cover in release notes.

### DIFF
--- a/entity-service-api/src/main/proto/org/hypertrace/entity/data/service/v1/entity_data_request.proto
+++ b/entity-service-api/src/main/proto/org/hypertrace/entity/data/service/v1/entity_data_request.proto
@@ -32,6 +32,8 @@ enum Operator {
   LE = 10;
   CONTAINS = 11;
   LIKE = 12;
+  EXISTS = 13;
+  NOT_EXISTS = 14;
 }
 
 message Query {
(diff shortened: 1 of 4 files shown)


inespo (dismissed):

tim-mwangi (dismissed):

inespo (approved):

kotharironak:
@inespo @tim-mwangi I have to fix the tests as data are not cleared between tests, and I have written tests with those assumptions. Re-written with appending UUID like any other test.

inespo (approved):

PR #63, opened 2020-12-23
discarding printing of full config to avoid any sensitive data leak

kotharironak:
As part of the issue - https://github.com/hypertrace/hypertrace/issues/137, where we identified that some sensitive information is logged (password for postgres), we are going with option (a) listed in the issue. There is no need to print `config` here.

### DIFF
--- a/entity-service/src/main/java/org/hypertrace/entity/service/EntityServiceConfig.java
+++ b/entity-service/src/main/java/org/hypertrace/entity/service/EntityServiceConfig.java
@@ -14,8 +14,6 @@ public class EntityServiceConfig {
   private final Config entityServiceConfig;
 
   public EntityServiceConfig(Config config) {
-
-    LOG.info(config.toString());
     entityServiceConfig = config.getConfig("entity-service");
     dataStoreType = entityServiceConfig.getString(DocumentStoreConfig.DATASTORE_TYPE_CONFIG_KEY);
   }


entity-service/src/main/java/org/hypertrace/entity/service/EntityServiceConfig.java
```diff
@@ -14,8 +14,6 @@
   private final Config entityServiceConfig;
 
   public EntityServiceConfig(Config config) {
-
-    LOG.info(config.toString());
```
kotharironak, line 18: This is anyway not required as we are already logging appConfig in the base framework - https://github.com/hypertrace/service-framework/blob/d3370d2309d95bc8fe7ca7275f66f351454422a6/platform-service-framework/src/main/java/org/hypertrace/core/serviceframework/PlatformService.java#L97

rish691 (approved):

entity-service/src/main/java/org/hypertrace/entity/service/EntityServiceConfig.java
buchi-busireddy: Won't the credentials be printed in the app config then?

kotharironak: Yes. And as discussed here - https://github.com/hypertrace/hypertrace/issues/137, I am going to apply option (a) even at the framework level too. The consumer should print requires things as per need with a meaningful message to make it actionable.  So, this is anyway needed.

kotharironak: This is the PR for the same - https://github.com/hypertrace/service-framework/pull/22, once that is merged, I have to upgrade appropriate services, at least attribute, entity and config-bootstrapper for now for postgres sensitive information.

JBAhire (approved):

tim-mwangi (approved):

Issue #56, opened 2020-12-08
Fixes streaming for large data 


inespo:
EntityDataService and EnttiyQueryService has an API to query a stream of entities, they are
` rpc query (Query) returns (stream Entity) {}` and `  rpc execute (EntityQueryRequest) returns (stream ResultSetChunk) {}` respectively.

The API is correct and this is important to support querying large entity result set. However, the implementation actually loads the entire document result set from document-store into memory before chunking them back into stream. 

If the query returns large number of entities for the client, this will result in OOM for entity-service.

Instead, the data from document should be chunked and streamed for each document or in batch until all of them are processed.

PR #65, opened 2020-12-24
Fix streaming large amount of data from entity-service

rish691:
## Description
https://github.com/hypertrace/entity-service/issues/56
The change is to correctly stream the data from the query apis, so that client doesn't face a data overload. 

### Testing
Added integration test

### DIFF
--- a/entity-service-impl/src/main/java/org/hypertrace/entity/data/service/DocumentParser.java
+++ b/entity-service-impl/src/main/java/org/hypertrace/entity/data/service/DocumentParser.java
@@ -10,7 +10,7 @@ import org.hypertrace.entity.service.util.DocStoreJsonFormat.Parser;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-class DocumentParser {
+public class DocumentParser {
   private static final Logger LOG = LoggerFactory.getLogger(DocumentParser.class);
   private static final Parser PARSER = DocStoreJsonFormat.parser().ignoringUnknownFields();
 
@@ -21,7 +21,7 @@ class DocumentParser {
     return (T) messageBuilder.build();
   }
 
-  <T extends Message> Optional<T> parseOrLog(
+  public <T extends Message> Optional<T> parseOrLog(
       @Nonnull Document document, @Nonnull Message.Builder messageBuilder) {
     try {
       return Optional.of(this.parseOrThrow(document, messageBuilder));
--- a/entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
+++ b/entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
@@ -20,6 +20,7 @@ import org.hypertrace.core.documentstore.Document;
 import org.hypertrace.core.documentstore.JSONDocument;
 import org.hypertrace.core.documentstore.SingleValueKey;
 import org.hypertrace.core.grpcutils.context.RequestContext;
+import org.hypertrace.entity.data.service.DocumentParser;
 import org.hypertrace.entity.data.service.v1.AttributeValue;
 import org.hypertrace.entity.data.service.v1.Entity;
 import org.hypertrace.entity.data.service.v1.Query;
@@ -49,6 +50,7 @@ public class EntityQueryServiceImpl extends EntityQueryServiceImplBase {
   private static final Logger LOG = LoggerFactory.getLogger(EntityQueryServiceImpl.class);
   private static final String ATTRIBUTE_MAP_CONFIG_PATH = "entity.service.attributeMap";
   private static final Parser PARSER = DocStoreJsonFormat.parser().ignoringUnknownFields();
+  private static final DocumentParser DOCUMENT_PARSER = new DocumentParser();
 
   private final Collection entitiesCollection;
   private final Map<String, Map<String, String>> attrNameToEDSAttrMap;
@@ -89,14 +91,30 @@ public class EntityQueryServiceImpl extends EntityQueryServiceImplBase {
         .convertToEDSQuery(request, attrNameToEDSAttrMap.get(request.getEntityType()));
     Iterator<Document> documentIterator = entitiesCollection.search(
         DocStoreConverter.transform(tenantId.get(), query));
-    List<Entity> entities = convertDocsToEntities(documentIterator);
 
-    //Build result
-    //TODO : chunk response. For now sending everything in one chunk
-    responseObserver.onNext(convertEntitiesToResultSetChunk(
-        entities,
-        request.getSelectionList(),
-        attrNameToEDSAttrMap.get(request.getEntityType())));
+    boolean isFirstEntity = true;
+    while (documentIterator.hasNext()) {
+      Optional<Entity> entity = DOCUMENT_PARSER.parseOrLog(documentIterator.next(), Entity.newBuilder());
+      if (entity.isPresent()) {
+        Row row = convertToEntityQueryResult(entity.get(), request.getSelectionList(), attrNameToEDSAttrMap.get(request.getEntityType()));
+        ResultSetChunk.Builder resultBuilder = ResultSetChunk.newBuilder();
+        // Set metadata for first entity
+        if (isFirstEntity) {
+          resultBuilder.setResultSetMetadata(ResultSetMetadata.newBuilder()
+              .addAllColumnMetadata(
+                  () -> request.getSelectionList().stream().map(Expression::getColumnIdentifier)
+                      .map(
+                          ColumnIdentifier::getColumnName)
+                      .map(s -> ColumnMetadata.newBuilder().setColumnName(s).build()).iterator())
+              .build());
+          isFirstEntity = false;
+        }
+        //Build data
+        resultBuilder.addRow(row);
+        responseObserver.onNext(resultBuilder.build());
+      }
+    }
+
     responseObserver.onCompleted();
   }
 
--- a/entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
+++ b/entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
@@ -93,7 +93,21 @@ public class EntityQueryServiceImpl extends EntityQueryServiceImplBase {
         DocStoreConverter.transform(tenantId.get(), query));
 
     boolean isFirstEntity = true;
-    int chunkId = 0;
+    int chunkId = 1;
+
+    if (!documentIterator.hasNext()) {
+      ResultSetChunk.Builder resultBuilder = ResultSetChunk.newBuilder();
+      resultBuilder.setResultSetMetadata(ResultSetMetadata.newBuilder()
+          .addAllColumnMetadata(
+              () -> request.getSelectionList().stream().map(Expression::getColumnIdentifier)
+                  .map(
+                      ColumnIdentifier::getColumnName)
+                  .map(s -> ColumnMetadata.newBuilder().setColumnName(s).build()).iterator())
+          .build());
+      resultBuilder.setIsLastChunk(true);
+      resultBuilder.setChunkId(chunkId);
+      responseObserver.onNext(resultBuilder.build());
+    }
     while (documentIterator.hasNext()) {
       Optional<Entity> entity = DOCUMENT_PARSER.parseOrLog(documentIterator.next(), Entity.newBuilder());
       ResultSetChunk.Builder resultBuilder = ResultSetChunk.newBuilder();
@@ -118,7 +132,7 @@ public class EntityQueryServiceImpl extends EntityQueryServiceImplBase {
         // if the optional is empty, subtly convey it to the consumer
         resultBuilder.setHasError(true);
       }
-      resultBuilder.setChunkId(++chunkId);
+      resultBuilder.setChunkId(chunkId++);
       resultBuilder.setIsLastChunk(!documentIterator.hasNext());
       responseObserver.onNext(resultBuilder.build());
     }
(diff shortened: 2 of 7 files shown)


entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -89,14 +91,29 @@ public void execute(EntityQueryRequest request, StreamObserver<ResultSetChunk> r
...
-        request.getSelectionList(),
-        attrNameToEDSAttrMap.get(request.getEntityType())));
+
+    boolean isFirstEntity = true;
+    while (documentIterator.hasNext()) {
+      Optional<Entity> entity = DOCUMENT_PARSER.parseOrLog(documentIterator.next(), Entity.newBuilder());
+      if (entity.isPresent()) {
+        Row row = convertToEntityQueryResult(entity.get(), request.getSelectionList(), attrNameToEDSAttrMap.get(request.getEntityType()));
+        ResultSetChunk.Builder resultBuilder = ResultSetChunk.newBuilder();
+        // Set metadata for first entity
```
buchi-busireddy, line 101: was the metadata not being sent earlier?

kotharironak: It was being sent. Currently, we were sending only one single chunk for all serarched entities.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -89,14 +91,29 @@ public void execute(EntityQueryRequest request, StreamObserver<ResultSetChunk> r
...
+              .addAllColumnMetadata(
+                  () -> request.getSelectionList().stream().map(Expression::getColumnIdentifier)
+                      .map(
+                          ColumnIdentifier::getColumnName)
+                      .map(s -> ColumnMetadata.newBuilder().setColumnName(s).build()).iterator())
+              .build());
+          isFirstEntity = false;
+        }
+        //Build data
+        resultBuilder.addRow(row);
```
kotharironak, line 113: Looks like, with chunking, we need to increment chunk_id and also need to take care of last chunk as per API - https://github.com/hypertrace/entity-service/blob/4025041f432818bdbfb0b6123cdd2fa7bedeadff/entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/response.proto

Query-service also seems to have a similar protocol (ResultSetChunk) - https://github.com/hypertrace/query-service/blob/466b05904139ff2f4c1de8de46597a03599d2ec3/query-service-api/src/main/proto/response.proto

You can also see the impl of chunk_id and is_last_chunk here as a reference - https://github.com/hypertrace/query-service/blob/466b05904139ff2f4c1de8de46597a03599d2ec3/query-service-impl/src/main/java/org/hypertrace/core/query/service/RowChunkingOperator.java

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -89,14 +91,29 @@ public void execute(EntityQueryRequest request, StreamObserver<ResultSetChunk> r
...
-    //Build result
-    //TODO : chunk response. For now sending everything in one chunk
-    responseObserver.onNext(convertEntitiesToResultSetChunk(
-        entities,
-        request.getSelectionList(),
-        attrNameToEDSAttrMap.get(request.getEntityType())));
+
+    boolean isFirstEntity = true;
+    while (documentIterator.hasNext()) {
+      Optional<Entity> entity = DOCUMENT_PARSER.parseOrLog(documentIterator.next(), Entity.newBuilder());
```
kotharironak, line 97: Earlier, we are using - `DocStoreJsonFormat` as part of https://github.com/hypertrace/entity-service/blob/4025041f432818bdbfb0b6123cdd2fa7bedeadff/entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java#L103

I think, this function would have been reused from `convertDocsToEntities` to `convertDocsToEntity`.
```
private Entity convertDocsToEntity(Document document) {
      Entity.Builder builder = Entity.newBuilder();
      try {
        PARSER.merge(next.toJson(), builder);
      } catch (InvalidProtocolBufferException e) {
        LOG.error("Could not deserialize the document into an entity.", e);
      }
    }
    return builder.build();
  }
```

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/DocumentParser.java
```diff
@@ -10,7 +10,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-class DocumentParser {
+public class DocumentParser {
```
kotharironak, line 13: Why is it to change to `public`?

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/DocumentParser.java
```diff
@@ -21,7 +21,7 @@
     return (T) messageBuilder.build();
   }
 
-  <T extends Message> Optional<T> parseOrLog(
+  public <T extends Message> Optional<T> parseOrLog(
```
kotharironak, line 24: why is this change to `public`?

entity-service/src/integrationTest/resources/configs/entity-service/application.conf
```diff
@@ -11,5 +11,20 @@ entity.service.config = {
...
+    "scope": "SERVICE",
+    "name": "SERVICE.id",
+    "subDocPath": "entityId"
+  },
+  {
+    "scope": "SERVICE",
+    "name": "SERVICE.name",
+    "subDocPath": "entityName"
+  },
+  {
```
kotharironak, line 25: you want to add one more attribute with `attributes.` as it has a special condition in code.
e.g `attributes.API_NAME`

entity-service-impl/src/test/java/org/hypertrace/entity/query/service/EntityQueryServiceImplTest.java
```diff
@@ -177,6 +197,106 @@ public void updateSuccess() throws Exception {
...
+            () -> {
+              EntityQueryServiceImpl eqs =
+                  new EntityQueryServiceImpl(mockEntitiesCollection, attributeFqnMaps);
+
+              eqs.execute(request, mockResponseObserver);
+              return null;
+            });
+
+    verify(mockEntitiesCollection, times(1)).search(any());
+    verify(mockResponseObserver, times(1)).onNext(any());
```
kotharironak, line 260: would consider adding one more document above, so we can check onNext called multiple times say ` times(2)` if we have two documents.
```
 List<Document> docs = java.util.Collections.singletonList(new JSONDocument(JsonFormat.printer().print(ENTITY)));
```

entity-service/src/integrationTest/java/org/hypertrace/entity/service/service/EntityQueryServiceTest.java
```diff
@@ -0,0 +1,166 @@
...
+/**
+ * Test for {@link org.hypertrace.entity.query.service.client.EntityQueryServiceClient}
+ */
+public class EntityQueryServiceTest {
+
+  private static EntityQueryServiceClient entityQueryServiceClient;
+  private static EntityDataServiceClient entityDataServiceClient;
+  private static final String TENANT_ID =
+      "__testTenant__" + EntityQueryServiceClient.class.getSimpleName();
+  private static final String TEST_ENTITY_TYPE_V2 = "TEST_ENTITY";
```
kotharironak, line 52: nit: looks like not used - `TEST_ENTITY_TYPE_V2`

kotharironak, line 52:
```suggestion
```

entity-service/src/integrationTest/java/org/hypertrace/entity/service/service/EntityQueryServiceTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+  @AfterAll
+  public static void teardown() {
+    IntegrationTestServerUtil.shutdownServices();
+  }
+
+  private static void setupEntityTypes(Channel channel) {
+    org.hypertrace.entity.type.service.client.EntityTypeServiceClient entityTypeServiceV1Client
+        = new org.hypertrace.entity.type.service.client.EntityTypeServiceClient(channel);
+    EntityTypeServiceClient entityTypeServiceV2Client = new EntityTypeServiceClient(channel);
```
kotharironak, line 73: Looks like not used - `entityTypeServiceV2Client`

kotharironak, line 73:
```suggestion
```

entity-service/src/integrationTest/java/org/hypertrace/entity/service/service/EntityQueryServiceTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+    Entity entity2 = Entity.newBuilder()
+        .setTenantId(TENANT_ID)
+        .setEntityType(EntityType.SERVICE.name())
+        .setEntityName("Some Service 2")
+        .putIdentifyingAttributes(
+            EntityConstants.getValue(CommonAttribute.COMMON_ATTRIBUTE_FQN),
+            generateRandomUUIDAttrValue())
+        .build();
+    Entity createdEntity2 = entityDataServiceClient.upsert(entity2);
```
kotharironak, line 110: Add assertion for this entity too - `assertNotNull(createdEntity2);`

entity-service/src/integrationTest/java/org/hypertrace/entity/service/service/EntityQueryServiceTest.java
```diff
@@ -0,0 +1,166 @@
...
+                            .build())
+                        .build()).
+                    build())
+                .build())
+        .addSelection(Expression.newBuilder().setColumnIdentifier(ColumnIdentifier.newBuilder().setColumnName("SERVICE.id").build()).build())
+        .addSelection(Expression.newBuilder().setColumnIdentifier(ColumnIdentifier.newBuilder().setColumnName("SERVICE.name").build()).build())
+        .build();
+
+    // this entity will be filtered out
+    Entity entity3 = Entity.newBuilder()
```
kotharironak, line 132: nit: can we add assertion that this entity created - `entity3` `assertNotNull(createdEntity3)`

kotharironak (commented):
overall looks good to me.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
rish691: Updated

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/DocumentParser.java
rish691: To be able to use the method `parseOrLog`

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
rish691: I am instead using `DocumentParser#parseOrLog`
Infact this method `convertDocsToEntities` should also be updated to use `DocumentParser#parseOrLog`. I will do it in the next pr.

entity-service/src/integrationTest/resources/configs/entity-service/application.conf
rish691: Its being tested in testEntityQueryServiceImplTest#testConvertToEntityQueryResult

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/DocumentParser.java
kotharironak: It was already used inside `searchByQueryAndStreamResponse` prior.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -89,14 +91,37 @@ public void execute(EntityQueryRequest request, StreamObserver<ResultSetChunk> r
...
-    //Build result
-    //TODO : chunk response. For now sending everything in one chunk
-    responseObserver.onNext(convertEntitiesToResultSetChunk(
-        entities,
-        request.getSelectionList(),
-        attrNameToEDSAttrMap.get(request.getEntityType())));
+
+    boolean isFirstEntity = true;
+    int chunkId = 0;
+    while (documentIterator.hasNext()) {
```
kotharironak, line 97: What if `documentIterator.hasNext()` is empty, what would be the return response? I think we were sending only metadata headers in ResultSetChunk but no rows if search results were empty - `entitiesCollection.search`. I think we should also do the same here.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
kotharironak: I see the reason you converted `DocumentParser#parseOrLog` to the public. But, what was the difference, and the reason you choose to move to use this function?

kotharironak: I see it is internally using the same PARSER.merge method.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
rish691: Makes sense, for backward compatibility of clients

kotharironak (approved):

PR #71, opened 2021-01-06
Revert "Fix streaming large amount of data from entity-service"

aaron-steinfeld:
Reverts hypertrace/entity-service#65

The change in chunking behavior breaks (bad) assumptions in gateway service. They need to be fixed before re-merging this.


skjindal93 (approved):

tim-mwangi (approved):

aaron-steinfeld:
CC @rish691 - not letting me add you as reviewer for whatever reason

jcchavezs:
Could you please point out those assumptions?

jcchavezs:
Also how can we avoid such things in the future?

aaron-steinfeld:
An example is here - https://github.com/hypertrace/gateway-service/blob/main/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityDataServiceEntityFetcher.java#L140
We're asking for result set metadata off each chunk - prior to #65 , since there was a single chunk, that was safe. After the change, only the first row will have metadata (since the documentation shows that only the first chunk contains metadata).

As to avoiding this specific thing:
- We already knew about this issue, and should have prioritized fixing it - but also could have been caught in review

As to the general case:
- Bringing in more integration tests across multiple services. 
- Nailing down the release process. The move away from continuous integration made this more difficult to track down, as an unrelated change later cut its own release which happened to include this change.
- Being more methodical and careful with API design to leave less room for assumptions would also be helpful. Having metadata on one chunk but another is just strange, IMO. 


Slightly orthogonal, but also want to bring up - the move to a single row per chunk, while not the explicit cause of this issue (other than in how it relates to the metadata), seems problematic. The purpose of the chunking is to support large data sets without a high transport overhead, so should contain multiple rows. The original implementation (at least in QS, where this was copied from) is 10K rows per chunk.

kotharironak:
In my opinion, it is hard to always catch such issues via code review as this is an example of a weak API assumption. In fact, during a review, by looking at this comment - https://github.com/hypertrace/entity-service/blob/main/entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/response.proto#L23, I had suggested returning metadata only on the first chuck.
So, I think, there needs to have an integration test for any assumed scenarios and not having assumptions on the API contract.

For this case, I can see two options  a) every chuck response should have metadata along with it b) indicate if metadata present in chunk response by updating the ResultSetChunk message
I would prefer a). Let me know what do you think guys? How were we thinking of resolving earlier by maintaining the assumption as part of the first chunk and changing consumer? 

cc @Rishbh: can we log this issue?

aaron-steinfeld:
> In my opinion, it is hard to always catch such issues via code review as this is an example of a weak API assumption. In fact, during a review, by looking at this comment

I think this is where it gets important to get reviews from consumers, the same as we would with an API change - we're changing the content of the result if not the structure. Not that this means we shouldn't have integration tests too.

I'm fine with returning metadata in every chunk, as long as we fix the implementation to return more data per chunk (see my response above about the single row per chunk implementation).

rish691:
> Slightly orthogonal, but also want to bring up - the move to a single row per chunk, while not the explicit cause of this issue (other than in how it relates to the metadata), seems problematic. The purpose of the chunking is to support large data sets without a high transport overhead, so should contain multiple rows. The original implementation (at least in QS, where this was copied from) is 10K rows per chunk.

I agree, I also wanted to return multiple rows per chunk, but I was not sure of what would be a good number to start with. I was wondering if it should be made configurable. 
How did we arrive at the number 10k?

aaron-steinfeld:
> How did we arrive at the number 10k?

Not sure how the number 10K was arrived at. Given that it was previously unlimited, it feels like a fine compromise though- a row of entity data is very little - assuming somewhere in the ballpark of 100B, we're saying around 1MB/chunk

> I was wondering if it should be made configurable.

The only place that would make sense to do that in my opinion would be adding it to the request (or metadata), since the client should be able to specify, not the server - but until we want to actually use a non-default value, no reason to add the complexity.

rish691:
> The only place that would make sense to do that in my opinion would be adding it to the request (or metadata), since the client should be able to specify, not the server - but until we want to actually use a non-default value, no reason to add the complexity.

Yeah, adding it in the server doesn't make sense, it should be passed in the request or I can go with default value for now.

aaron-steinfeld:
Let's do the default until we have a need to change it.

PR #72, opened 2021-01-07
Fix streaming large amount of data from entity-service

rish691:
## Description
* Stream the query execution result in EntityQueryService#execute
* Populate chunkId and set isLastChunk for ResultSetChunk
* Send metadata even when no search result
* Send multiple rows in a chunk

### Testing
Unit test and integration test

### DIFF
--- a/entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
+++ b/entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
@@ -9,6 +9,7 @@ import com.typesafe.config.Config;
 import io.grpc.stub.StreamObserver;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
@@ -95,8 +96,16 @@ public class EntityQueryServiceImpl extends EntityQueryServiceImplBase {
     //TODO: Optimize this later. For now converting to EDS Query and then again to DocStore Query.
     Query query = EntityQueryConverter
         .convertToEDSQuery(request, attrNameToEDSAttrMap.get(request.getEntityType()));
+    /**
+     * {@link EntityQueryRequest} selections need to treated differently, since they don't transform
+     * one to one to {@link org.hypertrace.entity.data.service.v1.EntityDataRequest} selections
+     */
+    Map<String, String> scopedAttrNameToEDSAttrMap = attrNameToEDSAttrMap.get(request.getEntityType());
+    List<String> docStoreSelections =
+        EntityQueryConverter.convertSelectionsToDocStoreSelections(
+            request.getSelectionList(), scopedAttrNameToEDSAttrMap);
     Iterator<Document> documentIterator = entitiesCollection.search(
-        DocStoreConverter.transform(tenantId.get(), query));
+        DocStoreConverter.transform(tenantId.get(), query, docStoreSelections));
 
     ResultSetMetadata resultSetMetadata = ResultSetMetadata.newBuilder()
         .addAllColumnMetadata(
@@ -241,8 +250,12 @@ public class EntityQueryServiceImpl extends EntityQueryServiceImplBase {
 
       // Finally return the selections
       Query entitiesQuery = Query.newBuilder().addAllEntityId(request.getEntityIdsList()).build();
-      Iterator<Document> documentIterator = entitiesCollection.search(
-          DocStoreConverter.transform(tenantId.get(), entitiesQuery));
+      List<String> docStoreSelections =
+          EntityQueryConverter.convertSelectionsToDocStoreSelections(
+              request.getSelectionList(), attributeFqnMap);
+      Iterator<Document> documentIterator =
+          entitiesCollection.search(
+              DocStoreConverter.transform(tenantId.get(), entitiesQuery, docStoreSelections));
       List<Entity> entities = convertDocsToEntities(documentIterator);
       responseObserver.onNext(convertEntitiesToResultSetChunk(
           entities,
--- a/entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
+++ b/entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
@@ -98,33 +98,28 @@ public class EntityQueryServiceImpl extends EntityQueryServiceImplBase {
     Iterator<Document> documentIterator = entitiesCollection.search(
         DocStoreConverter.transform(tenantId.get(), query));
 
+    ResultSetMetadata resultSetMetadata = ResultSetMetadata.newBuilder()
+        .addAllColumnMetadata(
+            () -> request.getSelectionList().stream().map(Expression::getColumnIdentifier)
+                .map(
+                    ColumnIdentifier::getColumnName)
+                .map(s -> ColumnMetadata.newBuilder().setColumnName(s).build()).iterator())
+        .build();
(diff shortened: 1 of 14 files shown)


entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -89,14 +97,54 @@ public void execute(EntityQueryRequest request, StreamObserver<ResultSetChunk> r
...
-    //Build result
-    //TODO : chunk response. For now sending everything in one chunk
-    responseObserver.onNext(convertEntitiesToResultSetChunk(
-        entities,
-        request.getSelectionList(),
-        attrNameToEDSAttrMap.get(request.getEntityType())));
+
+    if (!documentIterator.hasNext()) {
+      ResultSetChunk.Builder resultBuilder = ResultSetChunk.newBuilder();
+      resultBuilder.setResultSetMetadata(ResultSetMetadata.newBuilder()
```
aaron-steinfeld, line 103: Can we build the ResultSetMetadata once (it shouldn't change between chunks or cases) and avoid special casing the no data case? In case it's helpful, here's the QS implementation of chunking: https://github.com/hypertrace/query-service/blob/main/query-service-impl/src/main/java/org/hypertrace/core/query/service/RowChunkingOperator.java

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -89,14 +97,54 @@ public void execute(EntityQueryRequest request, StreamObserver<ResultSetChunk> r
...
+          .addAllColumnMetadata(
+              () -> request.getSelectionList().stream().map(Expression::getColumnIdentifier)
+                  .map(
+                      ColumnIdentifier::getColumnName)
+                  .map(s -> ColumnMetadata.newBuilder().setColumnName(s).build()).iterator())
+          .build());
+      resultBuilder.setIsLastChunk(true);
+      resultBuilder.setChunkId(1);
```
aaron-steinfeld, line 111: Id should start at 0 to match existing behavior (and on 115)

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -89,14 +97,54 @@ public void execute(EntityQueryRequest request, StreamObserver<ResultSetChunk> r
...
+        resultBuilder.setResultSetMetadata(ResultSetMetadata.newBuilder()
+            .addAllColumnMetadata(
+                () -> request.getSelectionList().stream().map(Expression::getColumnIdentifier)
+                    .map(
+                        ColumnIdentifier::getColumnName)
+                    .map(s -> ColumnMetadata.newBuilder().setColumnName(s).build()).iterator())
+            .build());
+        isNewChunk = false;
+      }
+      if (entity.isPresent()) {
```
aaron-steinfeld, line 130: style nit - `entity.ifPresent(entity => ...`

rish691: Skipping this, otherwise I'll need to make the variables used in lambda final.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
rish691: Building ResultSetMetadata once. 

I think the initial block is fine, I can get away with by moving to a `do while loop`, but that would be unnecessary.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
aaron-steinfeld: yes, that's another one of the benefits 😉

aaron-steinfeld (dismissed):

rish691:
@aaron-steinfeld before merging this one, let me run HT locally with these changes and check if everything is fine. 
Which screen broke last time?

aaron-steinfeld:
> @aaron-steinfeld before merging this one, let me run HT locally with these changes and check if everything is fine.
> Which screen broke last time?

Anywhere where we were fetching entity service data - try the various endpoint overview screens.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -89,14 +97,49 @@ public void execute(EntityQueryRequest request, StreamObserver<ResultSetChunk> r
...
+                .map(
+                    ColumnIdentifier::getColumnName)
+                .map(s -> ColumnMetadata.newBuilder().setColumnName(s).build()).iterator())
+        .build();
+    if (!documentIterator.hasNext()) {
+      ResultSetChunk.Builder resultBuilder = ResultSetChunk.newBuilder();
+      resultBuilder.setResultSetMetadata(resultSetMetadata);
+      resultBuilder.setIsLastChunk(true);
+      resultBuilder.setChunkId(0);
+      responseObserver.onNext(resultBuilder.build());
```
tim-mwangi, line 113: We can call onCompleted() and return here right? Seems confusing to have code below this since it implies that we should keep going(and I know there is `while (documentIterator.hasNext())` below but still confusing.)

rish691: yeah makes sense, let me update

aaron-steinfeld (approved):

aaron-steinfeld (approved):

skjindal93 (approved):
lgtm

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -86,26 +94,63 @@ public void execute(EntityQueryRequest request, StreamObserver<ResultSetChunk> r
...
     //TODO: Optimize this later. For now converting to EDS Query and then again to DocStore Query.
-    Map<String, String> scopedAttrNameToEDSAttrMap = attrNameToEDSAttrMap.get(request.getEntityType());
-    Query query = EntityQueryConverter.convertToEDSQuery(request, scopedAttrNameToEDSAttrMap);
+    Query query = EntityQueryConverter
+        .convertToEDSQuery(request, attrNameToEDSAttrMap.get(request.getEntityType()));
     /**
      * {@link EntityQueryRequest} selections need to treated differently, since they don't transform
      * one to one to {@link org.hypertrace.entity.data.service.v1.EntityDataRequest} selections
      */
+    Map<String, String> scopedAttrNameToEDSAttrMap = attrNameToEDSAttrMap.get(request.getEntityType());
```
skjindal93, line 103: nit: the map statement can be moved to top and can be re-used in the argument in above line

```
Map<String, String> scopedAttrNameToEDSAttrMap = attrNameToEDSAttrMap.get(request.getEntityType());
Query query = EntityQueryConverter.convertToEDSQuery(request, scopedAttrNameToEDSAttrMap);
```

kotharironak (approved):

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
rish691: Merging this one, will address it in a subsequent pr

PR #77, opened 2021-02-03
Update values.yaml

rish691:


helm/values.yaml
```diff
@@ -101,6 +101,10 @@ logConfig:
       enabled: false
 
 config-bootstrapper:
-  job:
-    prefix: entity
-  configurationCommands: {}
+  jobs:
```
tim-mwangi, line 104: Do you need to update the config bootstrapper dependency version somewhere?

jcchavezs:
@rish691 could you elaborate a bit on this change?

rish691:
> @rish691 could you elaborate a bit on this change?

@jcchavezs this change is no longer required. It was needed for this pr https://github.com/hypertrace/config-bootstrapper/pull/33 (which was closed in favour of https://github.com/hypertrace/config-bootstrapper/pull/34)

PR #83, opened 2021-02-11
fix:addresses not_in filter query for array type field - attributes.lables

kotharironak:
## Description
This PR addresses a gap for the handling of searching entities for not matching labels for array type field `attributes.labels` 
Currently, we have support for searching entities for matching labels, but we need similar capabilities to search entities for not matching labels. 

Currently, filtering entities for matching labels `attributes.labels IN (<val1> , <val2>)` is translated to `attributes.labels = <val1> OR attributes.labels = <val2> 
This PR adds support for `attributes.labels NOT_IN (<val1> , <val2>)` by translating it into `attributes.labels != <val1> AND attributes.labels != <val2>` 

### Testing
- added unit tests
- verified locally running MongoDB with the following data and query

Data:
```
db.mytest.insertOne(
   { item: "canvas", qty: 100, attributes: {labels: { valueList : {"values":[{"value":{"string":"attr-v1"}},{"value":{"string":"attr-v2"}},{"value":{"string":"attr-v3"}}]}} }}
)

db.mytest.insertOne(
   { item: "canvas", qty: 100, attributes: {labels: { valueList : {"values":[{"value":{"string":"attr-v01"}},{"value":{"string":"attr-v02"}},{"value":{"string":"attr-v03"}}]}} }}
)

db.mytest.insertOne(
   { item: "canvas", qty: 100, attributes: {labels: { valueList : {"values":[{"value":{"string":"attr-v01"}},{"value":{"string":"attr-v2"}},{"value":{"string":"attr-v4"}}]}} }}
)

db.mytest.insertOne(
   { item: "canvas", qty: 100, attributes: {labels: { valueList : {"values":[{"value":{"string":"attr-b1"}},{"value":{"string":"attr-b2"}},{"value":{"string":"attr-b4"}}]}} }}
)
```
Query:
```
db.mytest.find( { $and: [ {'attributes.labels.valueList.values': { $ne: {"value":{"string":"attr-b1"}}}}, {'attributes.labels.valueList.values': { $ne: {"value":{"string":"attr-v4"}}}}] } )
```

### Checklist:

### Documentation
None

### Note
We have a ticket for adding `NOT_IN` filter support in docstore interface - https://github.com/hypertrace/document-store/issues/32

### DIFF
--- a/entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
+++ b/entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
@@ -128,6 +128,8 @@ public class DocStoreConverter {
     if (filter.hasAttributeValue()) {
       if (ATTRIBUTES_LABELS_FIELD_NAME.equals(filter.getName()) && filter.getOperator() == Operator.EQ) {
         return transformToEqFilterWithValueListRhs(filter);
+      } else if (ATTRIBUTES_LABELS_FIELD_NAME.equals(filter.getName()) && filter.getOperator() == Operator.NEQ) {
+        return transformToNeqFilterWithValueListRhs(filter);
       } else if (ATTRIBUTES_LABELS_FIELD_NAME.equals(filter.getName()) && filter.getOperator() == Operator.IN) {
         return transformToOrFilterChainForStrArray(filter);
       } else if (ATTRIBUTES_LABELS_FIELD_NAME.equals(filter.getName()) && filter.getOperator() == Operator.NOT_IN) {
@@ -210,6 +212,32 @@ public class DocStoreConverter {
   private static Filter transformToEqFilterWithValueListRhs(AttributeFilter attributeFilter) {
     String fieldName = attributeFilter.getName() + VALUE_LIST_VALUES_CONST;
     return createEqFilterForAttributeValue(fieldName, attributeFilter.getAttributeValue());
+  }
+
+  private static Filter transformToNeqFilterWithValueListRhs(AttributeFilter attributeFilter) {
+    String fieldName = attributeFilter.getName() + VALUE_LIST_VALUES_CONST;
+
+    Filter f = new Filter();
+    f.setFieldName("");
+    f.setOp(Op.AND);
+
+    Filter [] childFilters = new Filter[2];
+
+    Filter exists = new Filter();
+    exists.setFieldName(fieldName);
+    exists.setOp(Op.EXISTS);
+    exists.setValue(true);
+    childFilters[0] = exists;
+
+    Filter neq = new Filter();
+    neq.setFieldName(fieldName);
+    neq.setOp(Op.NEQ);
+    neq.setValue(prepareRhsValueForSpecialValueListCase(attributeFilter.getAttributeValue()));
+    childFilters[1] = neq;
+
+    // Set child filters to empty array
+    f.setChildFilters(childFilters);
+    return f;
   }
 
   private static Filter createEqFilterForAttributeValue(String fieldName, AttributeValue attributeValue) {
--- a/entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
+++ b/entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
@@ -194,16 +194,10 @@ public class DocStoreConverter {
     f.setFieldName("");
     f.setOp(Op.AND);
 
-    Filter exists = new Filter();
-    exists.setFieldName(fieldName);
-    exists.setOp(Op.EXISTS);
-    exists.setValue(true);
-
     List<Filter> filters = attributeFilter.getAttributeValue().getValueList().getValuesList().stream()
         .map(rhsAttributeValue -> createNeqFilterForAttributeValue(fieldName, rhsAttributeValue))
         .collect(Collectors.toList());
 
-    filters.add(exists);
     f.setChildFilters(filters.toArray(new Filter[]{}));
 
     return f;
@@ -216,27 +210,12 @@ public class DocStoreConverter {
 
   private static Filter transformToNeqFilterWithValueListRhs(AttributeFilter attributeFilter) {
     String fieldName = attributeFilter.getName() + VALUE_LIST_VALUES_CONST;
-
     Filter f = new Filter();
-    f.setFieldName("");
-    f.setOp(Op.AND);
-
-    Filter [] childFilters = new Filter[2];
-
-    Filter exists = new Filter();
-    exists.setFieldName(fieldName);
-    exists.setOp(Op.EXISTS);
-    exists.setValue(true);
-    childFilters[0] = exists;
-
-    Filter neq = new Filter();
-    neq.setFieldName(fieldName);
-    neq.setOp(Op.NEQ);
-    neq.setValue(prepareRhsValueForSpecialValueListCase(attributeFilter.getAttributeValue()));
-    childFilters[1] = neq;
-
+    f.setFieldName(fieldName);
+    f.setOp(Op.NEQ);
+    f.setValue(prepareRhsValueForSpecialValueListCase(attributeFilter.getAttributeValue()));
     // Set child filters to empty array
-    f.setChildFilters(childFilters);
+    f.setChildFilters(new Filter[]{});
     return f;
   }
 
--- a/entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
+++ b/entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
@@ -192,12 +192,17 @@ public class DocStoreConverter {
     f.setFieldName("");
     f.setOp(Op.AND);
 
-    f.setChildFilters(
-        attributeFilter.getAttributeValue().getValueList().getValuesList().stream()
-            .map(rhsAttributeValue -> createNeqFilterForAttributeValue(fieldName, rhsAttributeValue))
-            .collect(Collectors.toList())
-            .toArray(new Filter[]{})
-    );
+    Filter exists = new Filter();
+    exists.setFieldName(fieldName);
+    exists.setOp(Op.EXISTS);
+    exists.setValue(true);
+
+    List<Filter> filters = attributeFilter.getAttributeValue().getValueList().getValuesList().stream()
+        .map(rhsAttributeValue -> createNeqFilterForAttributeValue(fieldName, rhsAttributeValue))
+        .collect(Collectors.toList());
+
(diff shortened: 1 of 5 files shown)


avinashkolluru (dismissed):
Looks fine. But is there no integration test for this?

avinashkolluru:
@kotharironak 
Does this return documents that dont even have the `attributes.labels` field?

kotharironak:
> Looks fine. But is there no integration test for this?

Yes. there is no integration test, but we can add. Let me do that.

entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
```diff
@@ -192,24 +211,36 @@ private static Filter createEqFilterForAttributeValue(String fieldName, Attribut
...
+    f.setValue(prepareRhsValueForSpecialValueListCase(attributeValue));
+    // Set child filters to empty array
+    f.setChildFilters(new Filter[]{});
+    return f;
+  }
+
+  private static Filter createNeqFilterForAttributeValue(String fieldName, AttributeValue attributeValue) {
+    Filter f = new Filter();
+    f.setFieldName(fieldName);
+    f.setOp(Op.NEQ);
```
skjindal93, line 223: Do we support NEQ operator in Mongo? I don't think we do

skjindal93 (changes requested):

entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
skjindal93: https://github.com/hypertrace/document-store/blob/main/document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java#L364

entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
```diff
@@ -183,6 +185,23 @@ private static Filter transformToOrFilterChainForStrArray(AttributeFilter attrib
     return f;
   }
 
+  private static Filter transformToAndFilterChainForStrArray(AttributeFilter attributeFilter) {
```
skjindal93, line 188: What about the fact the attribute itself might not exist? This would only work for the case, where the attribute exists

I think we would also want a NOT_EXISTS filter OR `transformToAndFilterChainForStrArray`

entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
```diff
@@ -130,6 +130,8 @@ private static Filter transform(AttributeFilter filter) {
         return transformToEqFilterWithValueListRhs(filter);
       } else if (ATTRIBUTES_LABELS_FIELD_NAME.equals(filter.getName()) && filter.getOperator() == Operator.IN) {
         return transformToOrFilterChainForStrArray(filter);
+      } else if (ATTRIBUTES_LABELS_FIELD_NAME.equals(filter.getName()) && filter.getOperator() == Operator.NOT_IN) {
```
skjindal93, line 133: Since, we are already adding support, can we also please add the support for 

```
if (ATTRIBUTES_LABELS_FIELD_NAME.equals(filter.getName()) && filter.getOperator() == Operator.NEQ) {
```

if possible?

entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
kotharironak: Yes, I was coming to that only. First, we have to implement NEQ. You had pending PR.

kotharironak: @skjindal93 https://github.com/hypertrace/document-store/pull/20

skjindal93: Yeah. I started the work, but then since there wasn't any requirement at that time, didn't finish it up. Feel free to use the make changes on the existing branch/PR :)

entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
kotharironak: Yes, based on `NEQ` impl here, we need `exits` condition in consumers. Added.

entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
kotharironak: we have merged the `NEQ` filter support PR for mongodb. So, this can go ahead. I have added integration test.

entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
kotharironak: Do we see this use case? Here, `EQ` means all the array elements matches, right? Or one of the values as well?

entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
```diff
@@ -183,6 +185,28 @@ private static Filter transformToOrFilterChainForStrArray(AttributeFilter attrib
...
+  private static Filter transformToAndFilterChainForStrArray(AttributeFilter attributeFilter) {
+    String fieldName = attributeFilter.getName() + VALUE_LIST_VALUES_CONST;
+
+    Filter f = new Filter();
+    f.setFieldName("");
+    f.setOp(Op.AND);
+
+    Filter exists = new Filter();
+    exists.setFieldName(fieldName);
+    exists.setOp(Op.EXISTS);
```
skjindal93, line 197: I am not exactly sure whether we need Exists or Not Exists here. Won't we want to return all the entities which do not have this field?

Here we are doing `labels NOT IN [A, B]`
- Do we only want to return entities which has labels but does not have A or B
- Or, do we also want to return entities where labels does not exist as well?

entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
skjindal93: > Here, EQ means all the array elements matches, right? Or one of the values as well?

I think it means any one of the values as well

entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
anandtiwary: I would think #2.

skjindal93: @kotharironak In that case, we need to do "NOT_EXISTS OR the other filter"

kotharironak: @skjindal93 Can you check this integration test - `testNotInFilterForMatchingLabelsQuery`? It has `5` documents. One doesn't have `attributes.labels` fields. Is it expected in the result?

skjindal93: > Can you check this integration test - testNotInFilterForMatchingLabelsQuery? It has 5 documents. One doesn't have attributes.labels fields. Is it expected in the result?

Yeah, `createdEntity5` is expected in the result

kotharironak: @skjindal93 I have added some more integration tests for `NOT_IN`, `IN`, and `EQ` - `testDifferentFilterForMatchingLabelsQuery`

entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
kotharironak: @skjindal93 I added support for this for completeness for `attributes.labels` But, all this logic may need change for Postgres support.

kotharironak:
> @kotharironak
> Does this return documents that dont even have the `attributes.labels` field?
> Thats why I feel an integration test will be helpful.

@avinashkolluru What is expected behavior here? The current implementation doesn't return. Check the test case `testDifferentFilterForMatchingLabelsQuery`, and it has one extra entity w/o having `attributes.lables` fields. Let me know the expected behavior, and accordingly, we can change.

entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
avinashkolluru: Yes @kotharironak That entity is expected in the result

kotharironak: Made the changes. @avinashkolluru @skjindal93, now it's good to go. Have a look.

entity-service/src/integrationTest/java/org/hypertrace/entity/service/service/EntityDataServiceTest.java
```diff
@@ -874,6 +875,167 @@ public void testEntityQueryOrderBy() {
...
+                .setAttributeValue(AttributeValue.newBuilder()
+                    .setValue(Value.newBuilder().setString("b1"))
+                )
+        ).build();
+
+    query = Query.newBuilder().setEntityType(EntityType.K8S_POD.name()).setFilter(attributeFilter)
+        .build();
+    entitiesList = entityDataServiceClient.query(TENANT_ID, query);
+    assertTrue(entitiesList.size() == 2);
+    assertTrue(entitiesList.contains(createdEntity2) && entitiesList.contains(createdEntity2));
```
skjindal93, line 980: I think there is a typo here 

```
entitiesList.contains(createdEntity2) && entitiesList.contains(createdEntity2)
```

skjindal93: I think it should be 

```
entitiesList.contains(createdEntity2) && entitiesList.contains(createdEntity5)
```

kotharironak: Done. Good catch, while debugging, I had checked `createdEntity5`, but had gone back & forth changing this condition.

skjindal93 (approved):

avinashkolluru (approved):

PR #84, opened 2021-02-11
refactor: allow extra identifying attributes on entity upsert

aaron-steinfeld:
## Description
This change relaxes the requirements on upsert so the identifying attributes of a V1-typed entity need not be an exact match of the defined identifying attributes for that type. As long as they're a supserset, the upsert is allowed. In other words, the defined identifying attributes are considered a minimum set.

This supports changes to add new identifying attributes to existing types without A. breaking existing IDs if new identifying attribute is not present and B. without breaking compatibility for existing upsert paths (which would otherwise require concurrent changes in the entity type config and the upsert path).

### Testing
Updated and ran unit tests. Ran integration tests

### Checklist:

### DIFF
--- a/entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityNormalizer.java
+++ b/entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityNormalizer.java
@@ -90,7 +90,7 @@ class EntityNormalizer {
               "No identifying attributes defined for EntityType: %s", request.getEntityType()));
     }
 
-    if (!idAttrNames.equals(request.getIdentifyingAttributesMap().keySet())) {
+    if (!request.getIdentifyingAttributesMap().keySet().containsAll(idAttrNames)) {
       throw new IllegalArgumentException(
           String.format(
               "Received and expected identifying attributes differ. Received: %s . Expected: %s",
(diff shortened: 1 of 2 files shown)


entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityNormalizer.java
```diff
@@ -90,7 +90,7 @@ private void verifyMatchingIdentifyingAttributes(String tenantId, Entity request
               "No identifying attributes defined for EntityType: %s", request.getEntityType()));
     }
 
-    if (!idAttrNames.equals(request.getIdentifyingAttributesMap().keySet())) {
+    if (!request.getIdentifyingAttributesMap().keySet().containsAll(idAttrNames)) {
```
avinashkolluru, line 93: This could potentially lead to multiple entities being created if an extra attribute(with high cardinality) is passed from the client.

aaron-steinfeld: If the client is passing an extra identifier, then that's presumably the intent, right?

avinashkolluru (approved):

PR #95, opened 2021-04-01
feat: add merge upsert api with condition support

aaron-steinfeld:
## Description
This change adds a new upsert API that has two main distinctions from the existing API:
1. Instead of overwriting an existing entity, it will merge with it
2. It supports a predicate that will be evaluated against any existing entity, to check before proceeding.

Both changes help support easier high write volume use cases - for example, multiple enrichers running in parallel. The merge case saves us from having to do two sequential round trips (to prevent overwriting data) in the update case, and the predicate helps us to better enforce write consistency - that is, ensuring a newer update doesn't arrive earlier than an older. Note this is not a full consistency guarantee, but likely good enough for our purposes.

### Testing
Ran E2E, also added ITs and UTs.

### Checklist:

### DIFF
--- a/entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java
+++ b/entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java
@@ -37,9 +37,12 @@ import org.hypertrace.entity.data.service.v1.EnrichedEntities;
 import org.hypertrace.entity.data.service.v1.EnrichedEntity;
 import org.hypertrace.entity.data.service.v1.Entities;
 import org.hypertrace.entity.data.service.v1.Entity;
+import org.hypertrace.entity.data.service.v1.Entity.Builder;
 import org.hypertrace.entity.data.service.v1.EntityDataServiceGrpc.EntityDataServiceImplBase;
 import org.hypertrace.entity.data.service.v1.EntityRelationship;
 import org.hypertrace.entity.data.service.v1.EntityRelationships;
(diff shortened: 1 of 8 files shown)


tim-mwangi (approved):

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java
```diff
@@ -496,6 +500,58 @@ public void getEnrichedEntityByTypeAndIdentifyingProps(
...
+      StreamObserver<MergeAndUpsertEntityResponse> responseObserver) {
+    RequestContext requestContext = RequestContext.CURRENT.get();
+    String tenantId = requestContext.getTenantId().orElse(null);
+    if (tenantId == null) {
+      responseObserver.onError(new ServiceException("Tenant id is missing in the request."));
+      return;
+    }
+
+    Entity receivedEntity = this.entityNormalizer.normalize(tenantId, request.getEntity());
+    Optional<Entity> existingEntity =
```
avinashkolluru, line 515: Can't we leverge https://github.com/hypertrace/document-store/blob/081cf1b5f9433e22d1a8736d01e1cc85dd663109/document-store/src/main/java/org/hypertrace/core/documentstore/Collection.java#L114 here?

aaron-steinfeld: This api doesn't support merging - here we need to fetch any existing entity. If none, create. If exists, evaluate condition and if condition succeeds, upsert. I mentioned the scenario on that PR, but dropped it since it seemed a little specific for ES.

PR #108, opened 2021-04-28
Add filtering support on multi value attributes

pavan-traceable:
## Description
Add generic support on filtering multi valued attributes from config

### Testing

### Checklist:

### Documentation

### DIFF
--- a/entity-service-api/src/main/proto/org/hypertrace/entity/data/service/v1/entity_data_request.proto
+++ b/entity-service-api/src/main/proto/org/hypertrace/entity/data/service/v1/entity_data_request.proto
@@ -59,6 +59,7 @@ message AttributeFilter {
   Operator operator = 2;
   AttributeValue attributeValue = 3;
   repeated AttributeFilter childFilter = 4;
+  bool is_multi_valued = 5;
 }
 
 message OrderByExpression {
--- a/entity-service-impl/build.gradle.kts
+++ b/entity-service-impl/build.gradle.kts
@@ -17,6 +17,7 @@ dependencies {
   implementation("com.github.f4b6a3:uuid-creator:3.5.0")
   implementation("io.reactivex.rxjava3:rxjava:3.0.11")
   implementation("com.google.guava:guava:30.1.1-jre")
+  implementation("org.apache.commons:commons-lang3:3.11")
 
   implementation("org.slf4j:slf4j-api:1.7.30")
   implementation("com.fasterxml.jackson.core:jackson-databind:2.12.2")
--- a/entity-service-impl/build.gradle.kts
+++ b/entity-service-impl/build.gradle.kts
@@ -7,6 +7,10 @@ plugins {
 dependencies {
   api(project(":entity-service-api"))
   api("org.hypertrace.core.serviceframework:service-framework-spi:0.1.23")
+
+  annotationProcessor("org.projectlombok:lombok:1.18.18")
+  compileOnly("org.projectlombok:lombok:1.18.18")
+
   implementation("org.hypertrace.core.documentstore:document-store:0.5.4")
   implementation("org.hypertrace.core.grpcutils:grpc-context-utils:0.4.0")
   implementation("org.hypertrace.core.grpcutils:grpc-client-utils:0.4.0")
@@ -17,7 +21,6 @@ dependencies {
   implementation("com.github.f4b6a3:uuid-creator:3.5.0")
   implementation("io.reactivex.rxjava3:rxjava:3.0.11")
   implementation("com.google.guava:guava:30.1.1-jre")
-  implementation("org.apache.commons:commons-lang3:3.11")
 
   implementation("org.slf4j:slf4j-api:1.7.30")
   implementation("com.fasterxml.jackson.core:jackson-databind:2.12.2")
--- a/entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityAttributeMapping.java
+++ b/entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityAttributeMapping.java
@@ -3,10 +3,8 @@ package org.hypertrace.entity.query.service;
 import static java.util.stream.Collectors.toUnmodifiableMap;
 
 import com.typesafe.config.Config;
-import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.stream.Collectors;
 import org.hypertrace.core.attribute.service.cachingclient.CachingAttributeClient;
 import org.hypertrace.core.attribute.service.v1.AttributeMetadata;
 import org.hypertrace.core.attribute.service.v1.AttributeSource;
@@ -24,7 +22,7 @@ class EntityAttributeMapping {
 
   private final CachingAttributeClient attributeClient;
   private final Map<String, String> explicitDocStoreMappingsByAttributeId;
-  private final List<String> multiValueAttributes;
+  private final Map<String, Boolean> multiValuedAttributeMap;
 
   EntityAttributeMapping(Config config, GrpcChannelRegistry channelRegistry) {
     this(
@@ -38,21 +36,20 @@ class EntityAttributeMapping {
                 toUnmodifiableMap(
                     conf -> conf.getString("name"), conf -> conf.getString(SUB_DOC_PATH))),
         config.getConfigList(ATTRIBUTE_MAP_CONFIG_PATH).stream()
-            .filter(
-                conf ->
-                    conf.hasPath(VALUE_KIND)
-                        && TYPE_STRING_ARRAY.equals(conf.getString(VALUE_KIND)))
-            .map(conf -> conf.getString(SUB_DOC_PATH))
-            .collect(Collectors.toUnmodifiableList()));
+            .filter(conf -> conf.hasPath(VALUE_KIND))
+            .collect(
+                toUnmodifiableMap(
+                    conf -> conf.getString("name"),
+                    conf -> TYPE_STRING_ARRAY.equals(conf.getString(VALUE_KIND)))));
   }
 
   EntityAttributeMapping(
       CachingAttributeClient attributeClient,
       Map<String, String> explicitDocStoreMappingsByAttributeId,
-      List<String> multiValueAttributes) {
+      Map<String, Boolean> multiValuedAttributeMap) {
     this.attributeClient = attributeClient;
     this.explicitDocStoreMappingsByAttributeId = explicitDocStoreMappingsByAttributeId;
-    this.multiValueAttributes = multiValueAttributes;
+    this.multiValuedAttributeMap = multiValuedAttributeMap;
   }
 
   /**
@@ -67,8 +64,11 @@ class EntityAttributeMapping {
         .or(() -> this.calculateDocStorePathFromAttributeId(requestContext, attributeId));
   }
 
-  public List<String> getMultiValuedAttributes() {
-    return multiValueAttributes;
+  public boolean isMultiValued(RequestContext requestContext, String attributeId) {
+    if (multiValuedAttributeMap.containsKey(attributeId)) {
+      return multiValuedAttributeMap.get(attributeId);
+    }
+    return getValueKindFromAttributeId(requestContext, attributeId);
   }
 
   private Optional<String> calculateDocStorePathFromAttributeId(
@@ -83,6 +83,19 @@ class EntityAttributeMapping {
                 .map(Optional::of)
                 .onErrorComplete()
                 .defaultIfEmpty(Optional.empty())
+                .blockingGet());
+  }
+
+  private boolean getValueKindFromAttributeId(RequestContext requestContext, String attributeId) {
+    return requestContext.call(
+        () ->
+            this.attributeClient
+                .get(attributeId)
+                .filter(metadata -> metadata.getSourcesList().contains(AttributeSource.EDS))
+                .map(AttributeMetadata::getValueKind)
+                .map(key -> TYPE_STRING_ARRAY.equals(key.name()))
+                .onErrorComplete()
+                .defaultIfEmpty(false)
                 .blockingGet());
   }
 }
--- a/entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityAttributeMapping.java
+++ b/entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityAttributeMapping.java
@@ -55,7 +55,16 @@ class EntityAttributeMapping {
   }
 
   public boolean isMultiValued(RequestContext requestContext, String attributeId) {
-    return getValueKindFromAttributeId(requestContext, attributeId);
+    return requestContext.call(
+        () ->
+            this.attributeClient
+                .get(attributeId)
+                .filter(metadata -> metadata.getSourcesList().contains(AttributeSource.EDS))
+                .map(AttributeMetadata::getValueKind)
+                .map(valueKind -> (AttributeKind.TYPE_STRING_ARRAY == valueKind))
+                .onErrorComplete()
+                .defaultIfEmpty(false)
+                .blockingGet());
   }
 
   private Optional<String> calculateDocStorePathFromAttributeId(
@@ -70,19 +79,6 @@ class EntityAttributeMapping {
                 .map(Optional::of)
                 .onErrorComplete()
                 .defaultIfEmpty(Optional.empty())
-                .blockingGet());
-  }
-
-  private boolean getValueKindFromAttributeId(RequestContext requestContext, String attributeId) {
-    return requestContext.call(
-        () ->
-            this.attributeClient
-                .get(attributeId)
-                .filter(metadata -> metadata.getSourcesList().contains(AttributeSource.EDS))
-                .map(AttributeMetadata::getValueKind)
-                .map(valueKind -> (AttributeKind.TYPE_STRING_ARRAY == valueKind))
-                .onErrorComplete()
-                .defaultIfEmpty(false)
                 .blockingGet());
   }
 }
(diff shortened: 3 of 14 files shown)


entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityAttributeMapping.java
```diff
@@ -52,6 +67,10 @@
         .or(() -> this.calculateDocStorePathFromAttributeId(requestContext, attributeId));
   }
 
+  public List<String> getMultiValuedAttributes() {
+    return multiValueAttributes;
```
aaron-steinfeld, line 71: Can we reverse this to something similar to above like `isMultiValued(RequestContext requestContext, String attributeId). The reason being this impl would only support v1-typed entities. V2 entities are not stored in the ES metadata, they're only available from attribute service (single source of truth) and we need to request them from there.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityAttributeMapping.java
```diff
@@ -30,14 +36,23 @@
         config.getConfigList(ATTRIBUTE_MAP_CONFIG_PATH).stream()
             .collect(
                 toUnmodifiableMap(
-                    conf -> conf.getString("name"), conf -> conf.getString("subDocPath"))));
+                    conf -> conf.getString("name"), conf -> conf.getString(SUB_DOC_PATH))),
+        config.getConfigList(ATTRIBUTE_MAP_CONFIG_PATH).stream()
+            .filter(
+                conf ->
+                    conf.hasPath(VALUE_KIND)
```
aaron-steinfeld, line 43: These can't be stored in a flat list - should be in a map keyed by scope since there may be duplicate keys with different types in different scopes.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityAttributeMapping.java
```diff
@@ -30,14 +36,23 @@
...
+                        && TYPE_STRING_ARRAY.equals(conf.getString(VALUE_KIND)))
+            .map(conf -> conf.getString(SUB_DOC_PATH))
+            .collect(Collectors.toUnmodifiableList()));
   }
 
   EntityAttributeMapping(
       CachingAttributeClient attributeClient,
-      Map<String, String> explicitDocStoreMappingsByAttributeId) {
+      Map<String, String> explicitDocStoreMappingsByAttributeId,
+      List<String> multiValueAttributes) {
```
skjindal93, line 52: Can we convert it in a Set, rather than List for O(1) search?

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityAttributeMapping.java
```diff
@@ -16,9 +18,13 @@
   private static final String ATTRIBUTE_SERVICE_HOST = "attribute.service.config.host";
   private static final String ATTRIBUTE_SERVICE_PORT = "attribute.service.config.port";
   static final String ENTITY_ATTRIBUTE_DOC_PREFIX = "attributes.";
+  public static final String TYPE_STRING_ARRAY = "TYPE_STRING_ARRAY";
+  public static final String VALUE_KIND = "valueKind";
+  public static final String SUB_DOC_PATH = "subDocPath";
 
   private final CachingAttributeClient attributeClient;
   private final Map<String, String> explicitDocStoreMappingsByAttributeId;
+  private final List<String> multiValueAttributes;
```
skjindal93, line 27: I think we call it `multiValuedAttributes` at some place, and `multiValueAttributes` at other places. Can we have a consistent naming scheme? Either is fine

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityAttributeMapping.java
avinashkolluru: Discussed a bit with Pavan. The `AttributeFilter` proto needs such a field for this to work in the current scheme of things.
Similar to how we convert the Attribute Id to the name field in AttributeFilter using the above method, we'll be doing the same for the isMultiValued field. Once the AttributeFilter has the isMultiValued field, the conversion to doc store query can be based on it.

aaron-steinfeld: Why does the requester need to provide metadata? The attribute id should be all we need

pavan-traceable: Can you please review and see if something can be changed ?

aaron-steinfeld: I had suggested above - `boolean isMultiValued(RequestContext requestContext, String attributeId)` - did that run into issues? That's identical to what we're doing for getting the doc store path, so all the infrastructure is in place.

pavan-traceable: I have added same.

avinashkolluru: @aaron-steinfeld thats exactly what Pavan did. But, that alone is not enough because the converter which converts from AttributeFilter to doc store filter is where this info is needed. Which means it needs to be available in the AttributeFilter proto or is there a simpler way?

aaron-steinfeld: Sorry, missed that there were new commits. Let me take a look

entity-service-impl/build.gradle.kts
```diff
@@ -17,6 +17,7 @@ dependencies {
   implementation("com.github.f4b6a3:uuid-creator:3.5.0")
   implementation("io.reactivex.rxjava3:rxjava:3.0.11")
   implementation("com.google.guava:guava:30.1.1-jre")
+  implementation("org.apache.commons:commons-lang3:3.11")
```
aaron-steinfeld, line 20: Not worth a second util lib just for pair.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityAttributeMapping.java
```diff
@@ -30,14 +34,22 @@
         config.getConfigList(ATTRIBUTE_MAP_CONFIG_PATH).stream()
             .collect(
                 toUnmodifiableMap(
-                    conf -> conf.getString("name"), conf -> conf.getString("subDocPath"))));
+                    conf -> conf.getString("name"), conf -> conf.getString(SUB_DOC_PATH))),
+        config.getConfigList(ATTRIBUTE_MAP_CONFIG_PATH).stream()
```
aaron-steinfeld, line 38: Actually, given that this is a new feature and has no compatibility requirements, do we even need to add this to config at all? Why not always retrieve from AS (the subdoc path from config we needed to continue supporting for backwards compat)

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityAttributeMapping.java
```diff
@@ -66,4 +85,17 @@
...
   }
+
+  private boolean getValueKindFromAttributeId(RequestContext requestContext, String attributeId) {
+    return requestContext.call(
+        () ->
+            this.attributeClient
+                .get(attributeId)
+                .filter(metadata -> metadata.getSourcesList().contains(AttributeSource.EDS))
+                .map(AttributeMetadata::getValueKind)
+                .map(valueKind -> TYPE_STRING_ARRAY.equals(valueKind.name()))
```
aaron-steinfeld, line 96: can we do an enum-enum comparison here?

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryConverter.java
```diff
@@ -336,20 +339,24 @@ private AttributeFilter convertToAttributeFilter(
     return builder;
   }
 
-  private String convertToAttributeKey(RequestContext requestContext, Expression expression) {
+  private Pair<String, Boolean> convertToAttributeKey(
```
aaron-steinfeld, line 342: so aside from my feelings about bringing in a second util lib, pair is generally a bad tool because it drops all semantics and it makes it harder for the reader to understand.`attributeData.getRight()` doesn't tell me what this is going to do, I have to read backwards to where that pair was generated to see what it holds. Much cleaner is too create a util class like

```java
  @Value
  private static class ColumnInformation {
    String columnName;
    boolean multiValued;
  }
```

now `builder.setIsMultiValued(columnInformation.isMultiValued())` and `String edsColumnName = columnInformation.geColumnName()` would be more readable.

Regardless, this function is no longer returning just a key and should be renamed.

entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
```diff
@@ -128,17 +128,13 @@ public static Filter getTenantIdEqFilter(String tenantId) {
 
   private static Filter transform(AttributeFilter filter) {
     if (filter.hasAttributeValue()) {
-      if (ATTRIBUTES_LABELS_FIELD_NAME.equals(filter.getName())
-          && filter.getOperator() == Operator.EQ) {
+      if (isMultiValuedAttribute(filter) && filter.getOperator() == Operator.EQ) {
```
aaron-steinfeld, line 131: OK, got the issue - because EQS goes through EDS (which has been one of our longest standing tech debts) the metadata gets lost.

So our options, are 
1. Fix that now and have EQS go directly to doc store
2. Find some way of reading metadata in EDS, but difficult since we're working in sub doc paths at this point
3. Pass this info along, for example via filter as implemented here.

I think everyone can guess my preference 😉 , but I get why it's easier to do 3 and I'm fine with it - but I do have a question - does anyone know was this tech debt was initially incurred? The translation from EQS->EDS looks just as involved as the translation from EDS->doc store, so what did we save by doing it?

avinashkolluru: I plead guilty for adding that tech debt, but honestly can't remember why I did that.

aaron-steinfeld: Certainly been in the same boat! It may be worth a 5m eval to just see if it would be hard to migrate now - IMO it's worth the extra work to avoid changing the EDS protos if we're talking hours and not days.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryConverter.java
aaron-steinfeld: yeah on the proto side the col name is already a top level thing, so not sure we can really change it cleanly.

pavan-traceable: I started with adding class, but there is no lombok here, so have to add all functions. And it will be used here only, so thought of going with Pair. I agree with readability part will add this class.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityAttributeMapping.java
pavan-traceable: Agree, apart from backward compatibility i thought it is added for performance concerns also.  There is client side caching in client, but i thought through config we want to limit calls to attribute service.

entity-service-impl/build.gradle.kts
pavan-traceable: Will remove

entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
pavan-traceable: With less context on Entity and Attribute Service and there is strict deadline for this feature from customer, don't want to do this at this point of time. Can someone of you please take the call ?

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityAttributeMapping.java
aaron-steinfeld: The cached calls shouldn't have any significant perf impacts - it would be 1 call per tenant per 15m, by default (and that cache is shared by the other part of the mapping on col name, anyway)

entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
aaron-steinfeld: This change would be internal to ES (which is why I like it vs the proto change which is external) and not involve AS - it would just be changing the conversion from EQS->EDS to an EQS->doc store and calling the doc store directly. I'm fine with either call - if you think it will significantly impact the time to deliver the feature, we don't have to address here.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityAttributeMapping.java
pavan-traceable: done. Thanks

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryConverter.java
pavan-traceable: done

entity-service-api/src/main/proto/org/hypertrace/entity/data/service/v1/entity_data_request.proto
```diff
@@ -59,6 +59,7 @@ message AttributeFilter {
   Operator operator = 2;
   AttributeValue attributeValue = 3;
   repeated AttributeFilter childFilter = 4;
+  bool is_multi_valued = 5;
```
skjindal93, line 62: We should first publish the proto API separately, fix the consumers to send the correct boolean flag, otherwise, merging this PR would mean that labels filter would start failing, because I am guessing `is_multi_valued` will be false by default

entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
```diff
@@ -128,17 +128,13 @@ public static Filter getTenantIdEqFilter(String tenantId) {
 
   private static Filter transform(AttributeFilter filter) {
     if (filter.hasAttributeValue()) {
-      if (ATTRIBUTES_LABELS_FIELD_NAME.equals(filter.getName())
```
skjindal93, line 131: Can we get rid of the `ATTRIBUTES_LABELS_FIELD_NAME` constant from the class?

entity-service-api/src/main/proto/org/hypertrace/entity/data/service/v1/entity_data_request.proto
pavan-traceable: is_multi_valued will be fetched from AttributeService for labels. Please check EntityAttributeMapping.

entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
pavan-traceable: done

entity-service-api/src/main/proto/org/hypertrace/entity/data/service/v1/entity_data_request.proto
skjindal93: I take that back, I realize since it's using attribute service, it should be fine, unless there are consumers of EDS which appliers on labels

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityAttributeMapping.java
```diff
@@ -52,6 +54,19 @@
...
+    return requestContext.call(
+        () ->
+            this.attributeClient
+                .get(attributeId)
+                .filter(metadata -> metadata.getSourcesList().contains(AttributeSource.EDS))
+                .map(AttributeMetadata::getValueKind)
+                .map(valueKind -> (AttributeKind.TYPE_STRING_ARRAY == valueKind))
+                .onErrorComplete()
+                .defaultIfEmpty(false)
+                .blockingGet());
```
skjindal93, line 67: I am not exactly sure, what's the correct norm, but shouldn't consumers of `isMultiValued` function call `blockingGet()`?

aaron-steinfeld: Generally, I prefer to leave things async by default and have the consumer decide (which is why attribute client is that way) but when I added the AS integration in this class, since it's a local impl class I chose to block so as to not introduce Rx across the service more broadly.

entity-service-impl/src/test/java/org/hypertrace/entity/query/service/EntityAttributeMappingTest.java
```diff
@@ -73,4 +76,41 @@ void returnsEmptyIfNoMapping() {
...
+        new EntityAttributeMapping(this.mockAttributeClient, Collections.emptyMap());
+    AttributeMetadata singleValueAttributeData =
+        AttributeMetadata.newBuilder()
+            .setKey("some-key")
+            .setValueKind(AttributeKind.TYPE_STRING)
+            .addSources(AttributeSource.EDS)
+            .build();
+    when(this.mockAttributeClient.get("some-id")).thenReturn(Single.just(singleValueAttributeData));
+
+    // Empty result, since the mock metadata doesn't have an entity source
```
aaron-steinfeld, line 93: this comment doesn't match the input (it has an entity source, but the wrong value kind)

aaron-steinfeld (commented):
lgtm; @tim-mwangi did you want to review since you did the initial impl here?

entity-service-impl/src/test/java/org/hypertrace/entity/query/service/EntityAttributeMappingTest.java
pavan-traceable: My bad, removed it

aaron-steinfeld (approved):

PR #112, opened 2021-07-20
fix: threading issue in entity caching client

aaron-steinfeld:
## Description
previously, a pending update could be retrieved and modified after it
has started updating, causing it to npe and leak memory

### DIFF
--- a/entity-data-service-rx-client/src/main/java/org/hypertrace/entity/data/service/rxclient/EntityDataCachingClient.java
+++ b/entity-data-service-rx-client/src/main/java/org/hypertrace/entity/data/service/rxclient/EntityDataCachingClient.java
@@ -136,6 +136,10 @@ class EntityDataCachingClient implements EntityDataClient {
         SingleObserver<Entity> observer,
         UpsertCondition condition,
         Duration maximumDelay) {
+      if (updateExecutionTimer.isDisposed()) {
+        throw new IllegalStateException("Attempting to add new update after execution");
+      }
+
       this.entityKey = entityKey;
       this.condition = condition;
       this.responseObservers.add(observer);
--- a/entity-data-service-rx-client/src/main/java/org/hypertrace/entity/data/service/rxclient/EntityDataCachingClient.java
+++ b/entity-data-service-rx-client/src/main/java/org/hypertrace/entity/data/service/rxclient/EntityDataCachingClient.java
@@ -6,6 +6,7 @@ import static java.util.Objects.nonNull;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
+import com.google.common.util.concurrent.Striped;
 import io.grpc.CallCredentials;
 import io.grpc.Channel;
 import io.reactivex.rxjava3.core.Completable;
@@ -21,6 +22,8 @@ import java.util.List;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReadWriteLock;
 import javax.annotation.Nonnull;
 import org.hypertrace.entity.data.service.v1.Entity;
 import org.hypertrace.entity.data.service.v1.EntityDataServiceGrpc;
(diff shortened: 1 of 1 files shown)


avinashkolluru (approved):

avinashkolluru (approved):

entity-data-service-rx-client/src/main/java/org/hypertrace/entity/data/service/rxclient/EntityDataCachingClient.java
```diff
@@ -112,9 +121,14 @@
...
-
-      EntityDataCachingClient.this.pendingEntityUpdates.remove(entityKey);
-
+      // Acquire write lock to ensure no more modification of this update
+      Lock lock = pendingUpdateStripedLock.get(entityKey).writeLock();
+      try {
+        lock.lock();
+        EntityDataCachingClient.this.pendingEntityUpdates.remove(entityKey);
+      } finally {
+        lock.unlock();
```
tim-mwangi, line 130: Is it ok to unlock immediately after removal or should we do the rest of the processing?

aaron-steinfeld: It should be - I initially had it on the whole method, but since the write lock is exclusive it seems like it'd be better to scope it. As soon as the pending update is removed from the map, future reads should be good to go as they'll trigger the creation of a new update.

entity-data-service-rx-client/src/main/java/org/hypertrace/entity/data/service/rxclient/EntityDataCachingClient.java
```diff
@@ -112,9 +121,14 @@
...
+      }
       Single<Entity> updateResult =
           EntityDataCachingClient.this.createOrUpdateEntity(entityKey, condition).cache();
       EntityDataCachingClient.this.cache.put(entityKey, updateResult);
```
satish-mittal, line 134: The above method at L133 `EntityDataCachingClient.this.createOrUpdateEntity(entityKey, condition).cache();` internally already adds the updated result into the cache:
```
  private Single<Entity> createOrUpdateEntity(
      EntityKey entityKey, UpsertCondition upsertCondition) {
    Single<Entity> updateResult =
        this.upsertEntityWithoutCaching(entityKey, upsertCondition).cache();
    EntityDataCachingClient.this.cache.put(entityKey, updateResult);
    return updateResult;
  }
```
Hence L134 is redundant.

entity-data-service-rx-client/src/main/java/org/hypertrace/entity/data/service/rxclient/EntityDataCachingClient.java
```diff
@@ -62,14 +68,17 @@
...
-      // but good hygiene
-      this.pendingEntityUpdates.put(entityKey, this.pendingEntityUpdates.get(entityKey));
+
+    // Acquire lock allowing multiple concurrent update adds, but only if no update executions
+    Lock lock = this.pendingUpdateStripedLock.get(entityKey).readLock();
+    try {
+      lock.lock();
+      this.pendingEntityUpdates
+          .computeIfAbsent(entityKey, unused -> new PendingEntityUpdate())
+          .addNewUpdate(entityKey, singleSubject, condition, maximumUpsertDelay);
```
satish-mittal, line 78: Trying to understand the code. Given the read lock, multiple threads will be allowed to go ahead and perform 2 operations concurrently:
1. Add a new entry within `pendingEntityUpdates` ConcurrentMap (if absent).
2. Mutate the `pendingEntityUpdate` instance fetched from the map as part of `addNewUpdate()`.

While step 1) is thread-safe given concurrentHashMap, step 2) is not thread-safe. Multiple threads may end up mutating `List<SingleObserver<Entity>> responseObservers` (since LinkedList is not thread-safe), along with overwriting other references like 
`private Disposable updateExecutionTimer;
    private Instant currentDeadline;
    private UpsertCondition condition;`

We need to synchronize the mutation of pendingEntityUpdate instance, right?

aaron-steinfeld: Yeah, that's certainly fair. the issues we were seeing were due to interactions between 1 + 2, but as you point out multiple `addNewUpdate` calls are not safe currently. Will follow up with another change in there, hopefully to support concurrency without locking.

aaron-steinfeld: Follow up in #114

entity-data-service-rx-client/src/main/java/org/hypertrace/entity/data/service/rxclient/EntityDataCachingClient.java
tim-mwangi: ok.

PR #114, opened 2021-07-21
fix: address more concurrency concerns in caching client

aaron-steinfeld:
## Description
Switching from a RW lock to mutual exclusion since adding updates is also not thread safe.

### DIFF
--- a/entity-data-service-rx-client/src/main/java/org/hypertrace/entity/data/service/rxclient/EntityDataCachingClient.java
+++ b/entity-data-service-rx-client/src/main/java/org/hypertrace/entity/data/service/rxclient/EntityDataCachingClient.java
@@ -130,7 +130,6 @@ class EntityDataCachingClient implements EntityDataClient {
       }
       Single<Entity> updateResult =
           EntityDataCachingClient.this.createOrUpdateEntity(entityKey, condition).cache();
-      EntityDataCachingClient.this.cache.put(entityKey, updateResult);
 
       responseObservers.forEach(updateResult::subscribe);
     }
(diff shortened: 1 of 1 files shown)


entity-data-service-rx-client/src/main/java/org/hypertrace/entity/data/service/rxclient/EntityDataCachingClient.java
```diff
@@ -121,8 +120,8 @@
     private UpsertCondition condition;
 
     private void executeUpdate() {
-      // Acquire write lock to ensure no more modification of this update
-      Lock lock = pendingUpdateStripedLock.get(entityKey).writeLock();
+      // Make sure no current additions
+      Lock lock = pendingUpdateStripedLock.get(entityKey);
```
satish-mittal, line 124: L133 seems redundant since cache update is already done within L132. Can be removed.

aaron-steinfeld: two different uses of the word cache.
```
Single<Entity> updateResult =
          EntityDataCachingClient.this.createOrUpdateEntity(entityKey, condition).cache();
Caches in the Single - that is, all subscribers will share a result rather than re-executing 
```
( http://reactivex.io/RxJava/3.x/javadoc/io/reactivex/rxjava3/core/Single.html#cache-- )
```
EntityDataCachingClient.this.cache.put(entityKey, updateResult);
```
puts the single itself into the guava cache for future callers

satish-mittal: Yes, but if we go inside:
```
private Single<Entity> createOrUpdateEntity(
      EntityKey entityKey, UpsertCondition upsertCondition) {
    Single<Entity> updateResult =
        this.upsertEntityWithoutCaching(entityKey, upsertCondition).cache();
    EntityDataCachingClient.this.cache.put(entityKey, updateResult);
    return updateResult;
  }
```
that is already adding the single within the guava cache, right?

aaron-steinfeld: Thanks for following up on this, my mind is clearly all over the place 🙁 - addressed.

satish-mittal (approved):

avinashkolluru (approved):

PR #121, opened 2021-08-02
impl for bulkupdate api

ankitchoudhary111:

### DIFF
--- a/entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
+++ b/entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
@@ -10,13 +10,17 @@ import io.grpc.stub.StreamObserver;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
+import org.hypertrace.core.documentstore.BulkUpdateResult;
 import org.hypertrace.core.documentstore.Collection;
 import org.hypertrace.core.documentstore.Datastore;
 import org.hypertrace.core.documentstore.Document;
 import org.hypertrace.core.documentstore.JSONDocument;
+import org.hypertrace.core.documentstore.Key;
 import org.hypertrace.core.documentstore.SingleValueKey;
 import org.hypertrace.core.grpcutils.client.GrpcChannelRegistry;
 import org.hypertrace.core.grpcutils.context.RequestContext;
@@ -24,6 +28,7 @@ import org.hypertrace.entity.data.service.DocumentParser;
 import org.hypertrace.entity.data.service.v1.AttributeValue;
 import org.hypertrace.entity.data.service.v1.Entity;
 import org.hypertrace.entity.data.service.v1.Query;
+import org.hypertrace.entity.query.service.v1.BulkEntityUpdateRequest;
 import org.hypertrace.entity.query.service.v1.ColumnIdentifier;
 import org.hypertrace.entity.query.service.v1.ColumnMetadata;
 import org.hypertrace.entity.query.service.v1.EntityQueryRequest;
@@ -37,6 +42,7 @@ import org.hypertrace.entity.query.service.v1.Row;
 import org.hypertrace.entity.query.service.v1.SetAttribute;
 import org.hypertrace.entity.query.service.v1.TotalEntitiesRequest;
 import org.hypertrace.entity.query.service.v1.TotalEntitiesResponse;
+import org.hypertrace.entity.query.service.v1.UpdateOperation;
 import org.hypertrace.entity.query.service.v1.Value;
 import org.hypertrace.entity.query.service.v1.ValueType;
 import org.hypertrace.entity.service.constants.EntityServiceConstants;
@@ -290,19 +296,115 @@ public class EntityQueryServiceImpl extends EntityQueryServiceImplBase {
           EntityQueryConverter.convertToAttributeValue(setAttribute.getValue()).build();
       String jsonValue = DocStoreJsonFormat.printer().print(attributeValue);
 
+      Map<Key, Map<String, Document>> toUpdate = new HashMap<>();
+
       for (String entityId : request.getEntityIdsList()) {
         SingleValueKey key =
             new SingleValueKey(requestContext.getTenantId().orElseThrow(), entityId);
         // TODO better error reporting once doc store exposes the,
-        if (!entitiesCollection.updateSubDoc(key, subDocPath, new JSONDocument(jsonValue))) {
-          LOG.warn(
-              "Failed to update entity {}, subDocPath {}, with new doc {}.",
-              key,
-              subDocPath,
-              jsonValue);
+        if (toUpdate.containsValue(key)) {
+          toUpdate.get(key).put(subDocPath, new JSONDocument(jsonValue));
+        } else {
+          Map<String, Document> subDocument = new HashMap<>();
+          subDocument.put(subDocPath, new JSONDocument(jsonValue));
+          toUpdate.put(key, subDocument);
         }
       }
+      try {
+        entitiesCollection.bulkUpdateSubDocs(toUpdate);
+      } catch (Exception e) {
+        LOG.warn(
+            "Failed to update entities, subDocPath {}, with new doc {}.", subDocPath, jsonValue);
+      }
     }
+  }
+
+  @Override
+  public void bulkUpdate(
+      BulkEntityUpdateRequest request, StreamObserver<ResultSetChunk> responseObserver) {
+    // Validations
+    RequestContext requestContext = RequestContext.CURRENT.get();
+    Optional<String> tenantId = requestContext.getTenantId();
+    if (tenantId.isEmpty()) {
+      responseObserver.onError(new ServiceException("Tenant id is missing in the request."));
+      return;
+    }
+    if (StringUtils.isEmpty(request.getEntityType())) {
+      responseObserver.onError(new ServiceException("Entity type is missing in the request."));
+      return;
+    }
+    if (request.getEntitiesCount() == 0) {
+      responseObserver.onError(new ServiceException("Entity IDs are missing in the request."));
+    }
+    try {
+      bulkDoUpdate(requestContext, request);
+      // Finally return the selections
+      List<String> entityIdsList = new ArrayList<>();
+      for (String entityId : request.getEntities().keySet()) {
+        entityIdsList.add(entityId);
+      }
+      Query entitiesQuery = Query.newBuilder().addAllEntityId(entityIdsList).build();
+      List<String> docStoreSelections =
+          entityQueryConverter.convertSelectionsToDocStoreSelections(
+              requestContext, request.getSelectionList());
+      Iterator<Document> documentIterator =
+          entitiesCollection.search(
+              DocStoreConverter.transform(tenantId.get(), entitiesQuery, docStoreSelections));
+      List<Entity> entities = convertDocsToEntities(documentIterator);
+      responseObserver.onNext(
+          convertEntitiesToResultSetChunk(requestContext, entities, request.getSelectionList()));
+      responseObserver.onCompleted();
+    } catch (Exception e) {
+      responseObserver.onError(
+          new ServiceException("Error occurred while executing " + request, e));
+    }
+  }
+
+  private void bulkDoUpdate(RequestContext requestContext, BulkEntityUpdateRequest request) {
+    Map<Key, Map<String, Document>> toUpdate = new HashMap<>();
+    Map<String, Document> documentMap = new HashMap<>();
+    for (String entityId : request.getEntities().keySet()) {
+      documentMap.clear();
+      try {
+        documentMap =
+            documentMapMaker(
+                request.getEntities().get(entityId).getUpdateOperationList(), requestContext);
+      } catch (Exception e) {
+        LOG.warn("Failed to update entity id {}", entityId);
+        continue;
+      }
+      if (!documentMap.isEmpty()) {
+        toUpdate.put(
+            new SingleValueKey(requestContext.getTenantId().orElseThrow(), entityId), documentMap);
+      }
+    }
+
+    try {
+      BulkUpdateResult bulkUpdateResult = entitiesCollection.bulkUpdateSubDocs(toUpdate);
+    } catch (Exception e) {
+      LOG.warn("Failed to update entities", e);
+    }
+  }
+
+  private Map<String, Document> documentMapMaker(
+      List<UpdateOperation> updateOperationList, RequestContext requestContext) throws IOException {
+    Map<String, Document> documentMap = new HashMap<>();
+    for (UpdateOperation updateOperation : updateOperationList) {
+      if (updateOperation.hasSetAttribute()) {
+        SetAttribute setAttribute = updateOperation.getSetAttribute();
+        String attributeId = setAttribute.getAttribute().getColumnName();
+        String subDocPath =
+            entityAttributeMapping
+                .getDocStorePathByAttributeId(requestContext, attributeId)
+                .orElseThrow(
+                    () -> new IllegalArgumentException("Unknown attribute FQN " + attributeId));
+        AttributeValue attributeValue =
+            EntityQueryConverter.convertToAttributeValue(setAttribute.getValue()).build();
+        String jsonValue = DocStoreJsonFormat.printer().print(attributeValue);
+        documentMap.put(subDocPath, new JSONDocument(jsonValue));
+      }
+    }
+    return documentMap;
   }
 
   @Override
--- a/entity-service-impl/src/test/java/org/hypertrace/entity/query/service/EntityQueryServiceImplTest.java
+++ b/entity-service-impl/src/test/java/org/hypertrace/entity/query/service/EntityQueryServiceImplTest.java
@@ -14,8 +14,10 @@ import com.google.protobuf.util.JsonFormat;
 import io.grpc.Context;
 import io.grpc.stub.StreamObserver;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.UUID;
+import org.hypertrace.core.documentstore.BulkUpdateResult;
 import org.hypertrace.core.documentstore.Collection;
 import org.hypertrace.core.documentstore.Document;
 import org.hypertrace.core.documentstore.Filter;
@@ -196,10 +198,13 @@ public class EntityQueryServiceImplTest {
             });
 
     verify(mockEntitiesCollection, times(1))
-        .updateSubDoc(
-            eq(new SingleValueKey("tenant1", "entity-id-1")),
-            eq("attributes.status"),
-            eq(new JSONDocument(DocStoreJsonFormat.printer().print(newStatus))));
+        .bulkUpdateSubDocs(
+            eq(
+                Map.of(
+                    new SingleValueKey("tenant1", "entity-id-1"),
+                    Map.of(
+                        "attributes.status",
+                        new JSONDocument(DocStoreJsonFormat.printer().print(newStatus))))));
   }
 
   @Test
@@ -485,9 +490,15 @@ public class EntityQueryServiceImplTest {
     }
   }
 
-  private Collection mockEntitiesCollection() {
+  private Collection mockEntitiesCollection() throws Exception {
     // mock successful update
-    return when(entitiesCollection.updateSubDoc(any(), any(), any())).thenReturn(true).getMock();
+    try {
+      return when(entitiesCollection.bulkUpdateSubDocs(any()))
+          .thenReturn(new BulkUpdateResult(0))
+          .getMock();
+    } catch (Exception e) {
+      throw e;
+    }
   }
 
   private RequestContext mockRequestContextWithTenantId() {
--- a/entity-service-impl/build.gradle.kts
+++ b/entity-service-impl/build.gradle.kts
@@ -11,7 +11,7 @@ dependencies {
   annotationProcessor("org.projectlombok:lombok:1.18.18")
   compileOnly("org.projectlombok:lombok:1.18.18")
 
-  implementation("org.hypertrace.core.documentstore:document-store:0.5.8")
+  implementation("org.hypertrace.core.documentstore:document-store:0.6.0")
   implementation("org.hypertrace.core.grpcutils:grpc-context-utils:0.4.0")
   implementation("org.hypertrace.core.grpcutils:grpc-client-utils:0.4.0")
   implementation("org.hypertrace.core.attribute.service:caching-attribute-service-client:0.12.3")
--- a/entity-service/build.gradle.kts
+++ b/entity-service/build.gradle.kts
@@ -18,7 +18,7 @@ dependencies {
   implementation("org.hypertrace.core.grpcutils:grpc-server-utils:0.4.0")
   implementation("org.hypertrace.core.grpcutils:grpc-client-utils:0.4.0")
   implementation("org.hypertrace.core.serviceframework:platform-service-framework:0.1.28")
-  implementation("org.hypertrace.core.documentstore:document-store:0.5.8")
+  implementation("org.hypertrace.core.documentstore:document-store:0.6.0")
 
   runtimeOnly("io.grpc:grpc-netty:1.36.1")
   constraints {


entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -290,19 +296,115 @@ private void doUpdate(RequestContext requestContext, EntityUpdateRequest request
...
+      responseObserver.onError(
+          new ServiceException("Error occurred while executing " + request, e));
+    }
+  }
+
+  private void bulkDoUpdate(RequestContext requestContext, BulkEntityUpdateRequest request) {
+    Map<Key, Map<String, Document>> toUpdate = new HashMap<>();
+    Map<String, Document> documentMap = new HashMap<>();
+    for (String entityId : request.getEntities().keySet()) {
+      documentMap.clear();
```
skjindal93, line 367: We should in fact create the `documentMap` inside the for loop, instead of clearing it always. Once you move it inside the for loop, can you also please rename it to `entityDocumentMap`?

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -290,19 +296,115 @@ private void doUpdate(RequestContext requestContext, EntityUpdateRequest request
...
+    } catch (Exception e) {
+      responseObserver.onError(
+          new ServiceException("Error occurred while executing " + request, e));
+    }
+  }
+
+  private void bulkDoUpdate(RequestContext requestContext, BulkEntityUpdateRequest request) {
+    Map<Key, Map<String, Document>> toUpdate = new HashMap<>();
+    Map<String, Document> documentMap = new HashMap<>();
+    for (String entityId : request.getEntities().keySet()) {
```
skjindal93, line 366: `request.getEntities()` is deprecated. Use `request.getEntitiesMap()`

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -290,19 +296,115 @@ private void doUpdate(RequestContext requestContext, EntityUpdateRequest request
...
+    if (tenantId.isEmpty()) {
+      responseObserver.onError(new ServiceException("Tenant id is missing in the request."));
+      return;
+    }
+    if (StringUtils.isEmpty(request.getEntityType())) {
+      responseObserver.onError(new ServiceException("Entity type is missing in the request."));
+      return;
+    }
+    if (request.getEntitiesCount() == 0) {
+      responseObserver.onError(new ServiceException("Entity IDs are missing in the request."));
```
skjindal93, line 337:
```suggestion
      responseObserver.onError(new ServiceException("Entities are missing in the request."));
```

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -290,19 +296,115 @@ private void doUpdate(RequestContext requestContext, EntityUpdateRequest request
...
+    }
+
+    try {
+      BulkUpdateResult bulkUpdateResult = entitiesCollection.bulkUpdateSubDocs(toUpdate);
+    } catch (Exception e) {
+      LOG.warn("Failed to update entities", e);
+    }
+  }
+
+  private Map<String, Document> documentMapMaker(
```
skjindal93, line 389: Please have a better descriptive name for the function. May be `transformUpdateOperations`

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
skjindal93: I don't think you even need the document map. You can just use the output of the `documentMapMaker` and directly put it inside the `toUpdate` map

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -290,19 +296,115 @@ private void doUpdate(RequestContext requestContext, EntityUpdateRequest request
...
+
+  private void bulkDoUpdate(RequestContext requestContext, BulkEntityUpdateRequest request) {
+    Map<Key, Map<String, Document>> toUpdate = new HashMap<>();
+    Map<String, Document> documentMap = new HashMap<>();
+    for (String entityId : request.getEntities().keySet()) {
+      documentMap.clear();
+      try {
+        documentMap =
+            documentMapMaker(
+                request.getEntities().get(entityId).getUpdateOperationList(), requestContext);
```
skjindal93, line 371:
```suggestion
                request.getEntitiesMap().get(entityId).getUpdateOperationList(), requestContext);
```

skjindal93: Since, we are getting the entities map multiple times, can you please rather get the entities map at the beginning of the function and use that, instead of `request.getEntitiesMap()` always

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -290,19 +296,115 @@ private void doUpdate(RequestContext requestContext, EntityUpdateRequest request
...
+      } catch (Exception e) {
```
skjindal93, line 372: We shouldn't have a try catch here. If there is an exception in transforming the update operations for an entity, we should fail the entire request

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -290,19 +296,115 @@ private void doUpdate(RequestContext requestContext, EntityUpdateRequest request
...
+      if (!documentMap.isEmpty()) {
+        toUpdate.put(
+            new SingleValueKey(requestContext.getTenantId().orElseThrow(), entityId), documentMap);
+      }
+    }
+
+    try {
+      BulkUpdateResult bulkUpdateResult = entitiesCollection.bulkUpdateSubDocs(toUpdate);
+    } catch (Exception e) {
+      LOG.warn("Failed to update entities", e);
```
skjindal93, line 385: Please log the entities map too

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -290,19 +296,115 @@ private void doUpdate(RequestContext requestContext, EntityUpdateRequest request
...
+    }
+  }
+
+  private Map<String, Document> documentMapMaker(
+      List<UpdateOperation> updateOperationList, RequestContext requestContext) throws IOException {
+    Map<String, Document> documentMap = new HashMap<>();
+    for (UpdateOperation updateOperation : updateOperationList) {
+      if (updateOperation.hasSetAttribute()) {
```
skjindal93, line 393: Can you please revert the condition?

```
if (!updateOperation.hasSetAttribute()) {
  continue;
}
```

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
skjindal93: Can this function be also used by `doUpdate` function? I think so

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -290,19 +296,115 @@ private void doUpdate(RequestContext requestContext, EntityUpdateRequest request
...
+                .getDocStorePathByAttributeId(requestContext, attributeId)
+                .orElseThrow(
+                    () -> new IllegalArgumentException("Unknown attribute FQN " + attributeId));
+        AttributeValue attributeValue =
+            EntityQueryConverter.convertToAttributeValue(setAttribute.getValue()).build();
+        String jsonValue = DocStoreJsonFormat.printer().print(attributeValue);
+        documentMap.put(subDocPath, new JSONDocument(jsonValue));
+      }
     }
+    return documentMap;
```
skjindal93, line 407:
```suggestion
    return Collections.unmodifiableMap(documentMap);
```

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -290,19 +296,115 @@ private void doUpdate(RequestContext requestContext, EntityUpdateRequest request
...
+        continue;
+      }
+      if (!documentMap.isEmpty()) {
+        toUpdate.put(
+            new SingleValueKey(requestContext.getTenantId().orElseThrow(), entityId), documentMap);
+      }
+    }
+
+    try {
+      BulkUpdateResult bulkUpdateResult = entitiesCollection.bulkUpdateSubDocs(toUpdate);
```
skjindal93, line 383: Can you please check if `toUpdate` is empty, and proceed only if non empty?

skjindal93: We also don't seem to be consuming `bulkUpdateResult` anywhere, so can get rid of that

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -290,19 +296,115 @@ private void doUpdate(RequestContext requestContext, EntityUpdateRequest request
...
+      responseObserver.onError(new ServiceException("Entity type is missing in the request."));
+      return;
+    }
+    if (request.getEntitiesCount() == 0) {
+      responseObserver.onError(new ServiceException("Entity IDs are missing in the request."));
+    }
+    try {
+      bulkDoUpdate(requestContext, request);
+      // Finally return the selections
+      List<String> entityIdsList = new ArrayList<>();
```
skjindal93, line 342:
```suggestion
      Set<String> entityIdsList = request.getEntitiesMap().keySet();
```

And, then can get rid of the for loop completely

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -290,19 +296,115 @@ private void doUpdate(RequestContext requestContext, EntityUpdateRequest request
...
+      Query entitiesQuery = Query.newBuilder().addAllEntityId(entityIdsList).build();
+      List<String> docStoreSelections =
+          entityQueryConverter.convertSelectionsToDocStoreSelections(
+              requestContext, request.getSelectionList());
+      Iterator<Document> documentIterator =
+          entitiesCollection.search(
+              DocStoreConverter.transform(tenantId.get(), entitiesQuery, docStoreSelections));
+      List<Entity> entities = convertDocsToEntities(documentIterator);
+      responseObserver.onNext(
+          convertEntitiesToResultSetChunk(requestContext, entities, request.getSelectionList()));
```
skjindal93, line 355: This seems to be duplicate code between `update` and `bulkUpdate` function. Can you please extract it out in a common function, along with the `convertEntitiesToResultSetChunk` function?

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -290,19 +291,128 @@ private void doUpdate(RequestContext requestContext, EntityUpdateRequest request
...
+      Set<String> entityIdsList = request.getEntitiesMap().keySet();
+      returnSelections(
+          entityIdsList, request.getSelectionList(), requestContext, tenantId, responseObserver);
+    } catch (Exception e) {
+      responseObserver.onError(
+          new ServiceException("Error occurred while executing " + request, e));
+    }
+  }
+
+  private void returnSelections(
```
skjindal93, line 347: We should always try to have a more descriptive name for the functions, something which the function is doing. For example here, the function is returning entities request from selections

```suggestion
  private void getProjectedEntities(
```

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -290,19 +291,128 @@ private void doUpdate(RequestContext requestContext, EntityUpdateRequest request
...
+      Iterable<String> entityIdsList,
+      List<Expression> selectionList,
+      RequestContext requestContext,
+      Optional<String> tenantId,
```
skjindal93, line 351: Can just send `String tenantId` and also annotate it with `@Nonnull`

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -290,19 +291,128 @@ private void doUpdate(RequestContext requestContext, EntityUpdateRequest request
...
+      StreamObserver<ResultSetChunk> responseObserver)
+      throws Exception {
+    Query entitiesQuery = Query.newBuilder().addAllEntityId(entityIdsList).build();
+    List<String> docStoreSelections =
+        entityQueryConverter.convertSelectionsToDocStoreSelections(requestContext, selectionList);
+    Iterator<Document> documentIterator =
+        entitiesCollection.search(
+            DocStoreConverter.transform(tenantId.get(), entitiesQuery, docStoreSelections));
+    List<Entity> entities = convertDocsToEntities(documentIterator);
+    responseObserver.onNext(
```
skjindal93, line 361: This shouldn't have the response observer at all. This should only return the list of entities. The consumers should take care of converting the entities into result set chunks and sending them via response observer

skjindal93:
There are some existing review comments. Please take care of them as well. Github must have hidden them.

skjindal93:
Once the review comments are addressed, please also add the unit and integration tests.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -290,19 +295,128 @@ private void doUpdate(RequestContext requestContext, EntityUpdateRequest request
...
+        LOG.warn("Failed to update entity id {}", entityId);
+        continue;
+      }
+      if (!documentMap.isEmpty()) {
+        entitiesUpdateMap.put(
+            new SingleValueKey(requestContext.getTenantId().orElseThrow(), entityId), documentMap);
+      }
+    }
+
+    if (entitiesUpdateMap.isEmpty()) {
```
aaron-steinfeld, line 389: shouldn't really be reachable - there should never be a successful case where we return without actually doing an update.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
aaron-steinfeld: So if the update fails, we catch it, log a warning and proceed on successfully? That doesn't seem right.

aaron-steinfeld (commented):
There's a lot of resolved comments here that aren't actually resolved. I tried to clear out my comments that had already been mentioned, but I may have repeated some stuff.

In the future @ankitchoudhary111 - generally the commenter, not the author, is responsible for resolving comments once they review with the author's fix.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -290,19 +295,128 @@ private void doUpdate(RequestContext requestContext, EntityUpdateRequest request
...
+        if (entitiesUpdateMap.containsKey(key)) {
+          entitiesUpdateMap.get(key).put(subDocPath, new JSONDocument(jsonValue));
+        } else {
+          Map<String, Document> subDocument = new HashMap<>();
+          subDocument.put(subDocPath, new JSONDocument(jsonValue));
+          entitiesUpdateMap.put(key, subDocument);
         }
       }
+      try {
+        entitiesCollection.bulkUpdateSubDocs(entitiesUpdateMap);
```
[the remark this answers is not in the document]
skjindal93: Couple of reasons

1. The update method currently only returns the number of successfully updated documents. https://github.com/hypertrace/document-store/blob/main/document-store/src/main/java/org/hypertrace/core/documentstore/Collection.java#L47

We just tried to follow what the existing `bulkUpdate` function does for entities https://github.com/hypertrace/document-store/blob/main/document-store/src/main/java/org/hypertrace/core/documentstore/Collection.java#L122

2. Trying to follow what the existing `update` API does. I think the use case was just to avoid another GET entities query to get the projected selections https://github.com/hypertrace/entity-service/blob/main/entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/entity_query_request.proto#L25

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -290,19 +295,128 @@ private void doUpdate(RequestContext requestContext, EntityUpdateRequest request
           EntityQueryConverter.convertToAttributeValue(setAttribute.getValue()).build();
       String jsonValue = DocStoreJsonFormat.printer().print(attributeValue);
 
+      Map<Key, Map<String, Document>> entitiesUpdateMap = new HashMap<>();
+
       for (String entityId : request.getEntityIdsList()) {
         SingleValueKey key =
             new SingleValueKey(requestContext.getTenantId().orElseThrow(), entityId);
         // TODO better error reporting once doc store exposes the,
```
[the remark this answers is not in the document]
skjindal93: I am not even sure, what the current comment even means, because it's incomplete

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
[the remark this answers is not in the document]
aaron-steinfeld: I'm really asking why `bulkUpdateSubDocs` doesn't, since it looks like at least here we need to do the following select/get anyway. Is it that the underlying store doesn't return this anyway, so even if we returned it from the api we're making two calls, an update and a select?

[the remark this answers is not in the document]
skjindal93: need to check the update support in the underlying store, but shouldn't the `bulkUpdateSubDocs` follow the same sematincs as `bulkUpdate`?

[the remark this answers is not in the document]
aaron-steinfeld: Yes, would make the same argument there - they both return the same object, so extending that object with the actual updated documents would affect both APIs.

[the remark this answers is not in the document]
skjindal93: Yeah, we could extend that to return the actual updated documents, but I don't think we would be able to get the projected entities with just the requested selections in that case, unless we actually pass the set of requested projections to the document store API, and then ask it to return the projected documents (_as part of the same request, if Mongo/Postgres can support that_)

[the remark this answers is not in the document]
skjindal93: In fact, these are 2 APIs merged into one
- bulk update entities
- get projected entities (_irrespective of the updated ones_)

now, that I think about it, the API sounds weird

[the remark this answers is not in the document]
aaron-steinfeld: So taking a glance at the impls, at least postgres (not sure about mongo) does not support this, so it would probably make sense to keep it as a distinct API - that decision should likely bubble up then so we're consistent - if entity service is forced to make two calls, then we may as well have the caller make the two calls since there are certainly cases where we don't care about the updated docs.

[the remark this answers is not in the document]
skjindal93: agreed. 

Mongo does support that, but I would rather prefer to drop the projections support from the API, as you mentioned too

@ankitchoudhary111 Can you please get rid of the projections support from the bulk update API?

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -289,20 +290,126 @@ private void doUpdate(RequestContext requestContext, EntityUpdateRequest request
       AttributeValue attributeValue =
           EntityQueryConverter.convertToAttributeValue(setAttribute.getValue()).build();
       String jsonValue = DocStoreJsonFormat.printer().print(attributeValue);
+      JSONDocument jsonDocument = new JSONDocument(jsonValue);
 
+      Map<Key, Map<String, Document>> entitiesUpdateMap = new HashMap<>();
       for (String entityId : request.getEntityIdsList()) {
         SingleValueKey key =
             new SingleValueKey(requestContext.getTenantId().orElseThrow(), entityId);
         // TODO better error reporting once doc store exposes the,
```
skjindal93, line 299: let's get rid of the TODO comment, since it's incomplete

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -289,20 +290,126 @@ private void doUpdate(RequestContext requestContext, EntityUpdateRequest request
...
+          Map<String, Document> subDocument = new HashMap<>();
+          subDocument.put(subDocPath, new JSONDocument(jsonValue));
+          entitiesUpdateMap.put(key, subDocument);
         }
       }
+      try {
+        entitiesCollection.bulkUpdateSubDocs(entitiesUpdateMap);
+      } catch (Exception e) {
+        LOG.error(
+            "Failed to update entities, subDocPath {}, with new doc {}.", subDocPath, jsonValue, e);
```
skjindal93, line 312:
```suggestion
            "Failed to update entities {}, subDocPath {}, with new doc {}.", entitiesUpdateMap, subDocPath, jsonValue, e);
```

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -289,20 +290,126 @@ private void doUpdate(RequestContext requestContext, EntityUpdateRequest request
...
             new SingleValueKey(requestContext.getTenantId().orElseThrow(), entityId);
         // TODO better error reporting once doc store exposes the,
-        if (!entitiesCollection.updateSubDoc(key, subDocPath, new JSONDocument(jsonValue))) {
-          LOG.warn(
-              "Failed to update entity {}, subDocPath {}, with new doc {}.",
-              key,
-              subDocPath,
-              jsonValue);
+        if (entitiesUpdateMap.containsKey(key)) {
+          entitiesUpdateMap.get(key).put(subDocPath, new JSONDocument(jsonValue));
```
skjindal93, line 301:
```suggestion
          entitiesUpdateMap.get(key).put(subDocPath, jsonDocument);
```

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -289,20 +290,126 @@ private void doUpdate(RequestContext requestContext, EntityUpdateRequest request
...
+        } else {
+          Map<String, Document> subDocument = new HashMap<>();
+          subDocument.put(subDocPath, new JSONDocument(jsonValue));
```
skjindal93, line 304:
```suggestion
          subDocument.put(subDocPath, jsonDocument);
```

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -289,20 +290,126 @@ private void doUpdate(RequestContext requestContext, EntityUpdateRequest request
...
+      return;
+    }
+    if (StringUtils.isEmpty(request.getEntityType())) {
+      responseObserver.onError(new ServiceException("Entity type is missing in the request."));
+      return;
+    }
+    if (request.getEntitiesCount() == 0) {
+      responseObserver.onError(new ServiceException("Entities are missing in the request."));
+    }
+    String tenantId = maybeTenantId.get();
```
skjindal93, line 335: Please pass in the `tenantId` directly to `doBulkUpdate`, instead of the `requestContext`

skjindal93: nit: an extra line above

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -289,20 +290,126 @@ private void doUpdate(RequestContext requestContext, EntityUpdateRequest request
       AttributeValue attributeValue =
           EntityQueryConverter.convertToAttributeValue(setAttribute.getValue()).build();
       String jsonValue = DocStoreJsonFormat.printer().print(attributeValue);
+      JSONDocument jsonDocument = new JSONDocument(jsonValue);
 
+      Map<Key, Map<String, Document>> entitiesUpdateMap = new HashMap<>();
```
skjindal93, line 295: I believe this entire logic can reuse the logic in `transformUpdateOperations`. 

Can you please see, if `transformUpdateOperations` can be re-usable both by `doUpdate` and `doBulkUpdate`? I think I did have a similar review comment before https://github.com/hypertrace/entity-service/pull/121#discussion_r682316361

skjindal93 (commented):
Please add integration tests https://github.com/hypertrace/entity-service/blob/main/entity-service/src/integrationTest/java/org/hypertrace/entity/service/service/EntityQueryServiceTest.java

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
ankitchoudhary111: doBulkUpdate calls the method transformUpdateOperations () which needs requestContext to find the subDocPath.
`String subDocPath =
          entityAttributeMapping
              .getDocStorePathByAttributeId(requestContext, attributeId)
              .orElseThrow(
                  () -> new IllegalArgumentException("Unknown attribute FQN " + attributeId));`

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
ankitchoudhary111: In `transformUpdateOperations` we are computing the `subDocPath ` and `jsonDocument.` 
The `Update` api only supports update operations where we update the same attribute with same value for multiple entities. So if we use `transformUpdateOperations` in `doUpdate` then we will unnecessarily compute `subDocPath` and `jsonDocument` again and again.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
aaron-steinfeld: Yeah, agree with keeping it request context - we really want to keep it as a context object until the last second, tenant IDs should only be resolved within the method that uses them.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -289,20 +289,127 @@ private void doUpdate(RequestContext requestContext, EntityUpdateRequest request
...
+              .getDocStorePathByAttributeId(requestContext, attributeId)
+              .orElseThrow(
+                  () -> new IllegalArgumentException("Unknown attribute FQN " + attributeId));
+      AttributeValue attributeValue =
+          EntityQueryConverter.convertToAttributeValue(setAttribute.getValue()).build();
+      try {
+        String jsonValue = DocStoreJsonFormat.printer().print(attributeValue);
+        documentMap.put(subDocPath, new JSONDocument(jsonValue));
+      } catch (Exception e) {
+        LOG.warn("Failed to put update corresponding to {} in the documentMap", subDocPath, e);
```
skjindal93, line 409: We should throw the exception back here

skjindal93:
```suggestion
        LOG.error("Failed to put update corresponding to {} in the documentMap", subDocPath, e);
```

entity-service-impl/src/test/java/org/hypertrace/entity/query/service/EntityQueryServiceImplTest.java
```diff
@@ -196,10 +200,143 @@ public void testUpdate_success() throws Exception {
...
+            eq(
+                Map.of(
+                    new SingleValueKey("tenant1", "entity-id-1"),
+                    Map.of(
+                        "attributes.status",
+                        new JSONDocument(DocStoreJsonFormat.printer().print(newStatus))))));
+  }
+
+  @Test
+  public void testBulkUpdate_noTenantId() throws Exception {
```
skjindal93, line 213: Can you put all the bulk update tests inside a nested test class?

entity-service-impl/src/test/java/org/hypertrace/entity/query/service/EntityQueryServiceImplTest.java
```diff
@@ -196,10 +200,143 @@ public void testUpdate_success() throws Exception {
...
+    Context.current()
+        .withValue(RequestContext.CURRENT, mockRequestContextWithTenantId())
+        .call(
+            () -> {
+              EntityQueryServiceImpl eqs =
+                  new EntityQueryServiceImpl(entitiesCollection, mockAttributeMapping, 1);
+              eqs.bulkUpdate(bulkUpdateRequest, mockResponseObserver);
+              return null;
+            });
+    verify(entitiesCollection, times(0)).bulkUpdateSubDocs(any());
```
skjindal93, line 295:
```suggestion
    verify(entitiesCollection, never()).bulkUpdateSubDocs(any());
```

entity-service/src/integrationTest/java/org/hypertrace/entity/service/service/EntityQueryServiceTest.java
```diff
@@ -68,7 +76,7 @@
   private static final Logger LOG = LoggerFactory.getLogger(EntityQueryServiceTest.class);
   private static final Slf4jLogConsumer logConsumer = new Slf4jLogConsumer(LOG);
 
-  private static EntityQueryServiceClient entityQueryServiceClient;
+  private static EntityQueryServiceBlockingStub blockingStub;
```
skjindal93, line 79:
```suggestion
  private static EntityQueryServiceBlockingStub entityQueryServiceClient;
```

entity-service/src/integrationTest/java/org/hypertrace/entity/service/service/EntityQueryServiceTest.java
```diff
@@ -82,6 +90,7 @@
   private static final String SERVICE_ID = generateRandomUUID();
 
   private static Map<String, String> apiAttributesMap;
+  private static Map<String, String> headers = Map.of("x-tenant-id", TENANT_ID);
```
skjindal93, line 93: statis final variables are upper cased 

```suggestion
  private static final Map<String, String> HEADERS = Map.of("x-tenant-id", TENANT_ID);
```

entity-service/src/integrationTest/java/org/hypertrace/entity/service/service/EntityQueryServiceTest.java
```diff
@@ -461,6 +477,122 @@ public void testCreateAndGetEntities() {
...
+            .setEntityName("api1")
+            .putIdentifyingAttributes(
+                EntityConstants.getValue(ServiceAttribute.SERVICE_ATTRIBUTE_ID),
+                createAttribute(SERVICE_ID))
+            .putIdentifyingAttributes(
+                EntityConstants.getValue(ApiAttribute.API_ATTRIBUTE_NAME), createAttribute("api1"))
+            .putIdentifyingAttributes(
+                EntityConstants.getValue(ApiAttribute.API_ATTRIBUTE_API_TYPE),
+                createAttribute(API_TYPE));
+    ;
```
skjindal93, line 495: dangling semi colon

entity-service/src/integrationTest/java/org/hypertrace/entity/service/service/EntityQueryServiceTest.java
```diff
@@ -461,6 +477,122 @@ public void testCreateAndGetEntities() {
...
+            .setEntityName("api2")
+            .putIdentifyingAttributes(
+                EntityConstants.getValue(ServiceAttribute.SERVICE_ATTRIBUTE_ID),
+                createAttribute(SERVICE_ID))
+            .putIdentifyingAttributes(
+                EntityConstants.getValue(ApiAttribute.API_ATTRIBUTE_NAME), createAttribute("api2"))
+            .putIdentifyingAttributes(
+                EntityConstants.getValue(ApiAttribute.API_ATTRIBUTE_API_TYPE),
+                createAttribute(API_TYPE));
+    ;
```
skjindal93, line 515: dangling semi colon

entity-service/src/integrationTest/java/org/hypertrace/entity/service/service/EntityQueryServiceTest.java
```diff
@@ -461,6 +477,122 @@ public void testCreateAndGetEntities() {
...
+            .addUpdateOperation(update2)
+            .build();
+    BulkEntityUpdateRequest bulkUpdateRequest =
+        BulkEntityUpdateRequest.newBuilder()
+            .setEntityType(EntityType.API.name())
+            .putEntities(entity1.getEntityId(), updateInfo1)
+            .putEntities(entity2.getEntityId(), updateInfo2)
+            .build();
+
+    Map<String, String> headers = Map.of("x-tenant-id", TENANT_ID);
```
skjindal93, line 560: We already have defined the headers at the class level. Can directly use them

entity-service-impl/src/test/java/org/hypertrace/entity/query/service/EntityQueryServiceImplTest.java
```diff
@@ -196,10 +201,149 @@ public void testUpdate_success() throws Exception {
...
+            eq(
+                Map.of(
+                    new SingleValueKey("tenant1", "entity-id-1"),
+                    Map.of(
+                        "attributes.status",
+                        new JSONDocument(DocStoreJsonFormat.printer().print(newStatus))))));
+  }
+
+  @Nested
+  class testBulkUpdate {
```
skjindal93, line 214: Class names are camel cased, starts with an upper case letter

```suggestion
  class BulkUpdateEntities {
```

entity-service/src/integrationTest/java/org/hypertrace/entity/service/service/EntityQueryServiceTest.java
```diff
@@ -82,6 +90,7 @@
   private static final String SERVICE_ID = generateRandomUUID();
 
   private static Map<String, String> apiAttributesMap;
+  private static Map<String, String> HEADERS = Map.of("x-tenant-id", TENANT_ID);
```
skjindal93, line 93:
```suggestion
  private static final Map<String, String> HEADERS = Map.of("x-tenant-id", TENANT_ID);
```

entity-service/src/integrationTest/java/org/hypertrace/entity/service/service/EntityQueryServiceTest.java
```diff
@@ -461,6 +477,119 @@ public void testCreateAndGetEntities() {
...
+            .addUpdateOperation(update2)
+            .build();
+    BulkEntityUpdateRequest bulkUpdateRequest =
+        BulkEntityUpdateRequest.newBuilder()
+            .setEntityType(EntityType.API.name())
+            .putEntities(entity1.getEntityId(), updateInfo1)
+            .putEntities(entity2.getEntityId(), updateInfo2)
+            .build();
+
+    Iterator<ResultSetChunk> resultChunkIterator =
```
skjindal93, line 558: Since, we aren't using `resultChunkIterator`, can we get rid of this?

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -289,20 +289,127 @@ private void doUpdate(RequestContext requestContext, EntityUpdateRequest request
...
+              .getDocStorePathByAttributeId(requestContext, attributeId)
+              .orElseThrow(
+                  () -> new IllegalArgumentException("Unknown attribute FQN " + attributeId));
+      AttributeValue attributeValue =
+          EntityQueryConverter.convertToAttributeValue(setAttribute.getValue()).build();
+      try {
+        String jsonValue = DocStoreJsonFormat.printer().print(attributeValue);
+        documentMap.put(subDocPath, new JSONDocument(jsonValue));
+      } catch (Exception e) {
+        LOG.error("Failed to put update corresponding to {} in the documentMap", subDocPath, e);
```
skjindal93, line 409: Please throw the exception back https://github.com/hypertrace/entity-service/pull/121#discussion_r685744920

skjindal93 (approved):

PR #123, opened 2021-08-09
fix: change entity client to not return single

aaron-steinfeld:
## Description
The motivation here is that the previous client API had major performance issues. Given the purpose of this api, high traffic writes, reducing to a simpler API makes sense for the safe of perf.

There were two major issues previously:
1. Each call returned a single, which requires a new subscription for each invocation. The subscriptions aren't _too_ heavy, but that's still a lot of unnecessary churn considering we don't actually read the result anywhere today.
2. By creating a subscription, the entire observable pipeline, including its lambda captures are retained in memory until the subscription is resolved (by default, 15s). This means that we're always holding 15s of ingestion data for no good reason. This needs to be fixed on the caller side, since the client can't control the context it's called in.

This was done as a breaking change, since these only currently have one consumer (enrichment) and that consumer needs to migrate.

### Testing
Updated UTs, compiled and tested full system locally.

### Checklist:

### DIFF
--- a/entity-data-service-rx-client/src/main/java/org/hypertrace/entity/data/service/rxclient/EntityDataCachingClient.java
+++ b/entity-data-service-rx-client/src/main/java/org/hypertrace/entity/data/service/rxclient/EntityDataCachingClient.java
@@ -11,19 +11,17 @@ import io.grpc.CallCredentials;
 import io.grpc.Channel;
 import io.reactivex.rxjava3.core.Completable;
 import io.reactivex.rxjava3.core.Single;
-import io.reactivex.rxjava3.core.SingleObserver;
 import io.reactivex.rxjava3.disposables.Disposable;
-import io.reactivex.rxjava3.subjects.SingleSubject;
 import java.time.Clock;
 import java.time.Duration;
 import java.time.Instant;
-import java.util.LinkedList;
-import java.util.List;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Lock;
 import javax.annotation.Nonnull;
+import lombok.extern.slf4j.Slf4j;
+import org.hypertrace.core.grpcutils.context.RequestContext;
 import org.hypertrace.entity.data.service.v1.Entity;
 import org.hypertrace.entity.data.service.v1.EntityDataServiceGrpc;
 import org.hypertrace.entity.data.service.v1.EntityDataServiceGrpc.EntityDataServiceStub;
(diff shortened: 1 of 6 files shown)


entity-data-service-rx-client/src/main/java/org/hypertrace/entity/data/service/rxclient/EntityDataCachingClient.java
```diff
@@ -128,24 +122,20 @@ private void executeUpdate() {
       } finally {
         lock.unlock();
       }
-      Single<Entity> updateResult =
-          EntityDataCachingClient.this.createOrUpdateEntity(entityKey, condition).cache();
-
-      responseObservers.forEach(updateResult::subscribe);
+      EntityDataCachingClient.this
+          .createOrUpdateEntity(entityKey, condition)
+          .doOnError(error -> log.error("Error upserting entity", error))
```
aaron-steinfeld, line 127: one downside of the void return is the error will never make it back - doesn't seem to me like it's worth creating a subscription per call though.

skjindal93: Can we still throw the exception back, if there was an exception creating/updating the entity?

avinashkolluru:
lgtm

entity-data-service-rx-client/src/main/java/org/hypertrace/entity/data/service/rxclient/EntityDataCachingClient.java
aaron-steinfeld: no, we've taken off the return of this method, and since it's not blocking this code is executed long after it returns. Abstractly, I prefer the alternative, which I had working - returning a single. But practically, that means that we'll be maintaining a subscription for each call/span for that "eventually" duration (which is currently 15s), just to turn around and take this same action.

skjindal93 (dismissed):

entity-data-service-rx-client/src/main/java/org/hypertrace/entity/data/service/rxclient/EntityDataCachingClient.java
aaron-steinfeld: This did remind me though, I need to break the habit of using `.doOnError` like this - since it doesn't catch it or subscribe to it. The logging still happens, but it'd be followed by a
```
io.reactivex.rxjava3.exceptions.UndeliverableException
The exception could not be delivered to the consumer because it has already canceled/disposed the flow or the exception has nowhere to go to begin with. Further reading: https://github.com/ReactiveX/RxJava/wiki/What's-different-in-2.0#error-handling | 
```
The following commit illustrates how this is supposed to be handled.

skjindal93 (approved):

tim-mwangi (approved):

PR #127, opened 2021-09-02
fix(entity-query-service): revert bulk update 

skjindal93:
Bulk update APIs are currently leading to corrupt entity documents. Reverting it back, till we fix the root cause

### DIFF
--- a/entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
+++ b/entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
@@ -7,18 +7,16 @@ import com.google.protobuf.InvalidProtocolBufferException;
 import com.google.protobuf.ServiceException;
 import com.typesafe.config.Config;
 import io.grpc.stub.StreamObserver;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
-import java.util.Map;
 import java.util.Optional;
 import org.hypertrace.core.documentstore.Collection;
 import org.hypertrace.core.documentstore.Datastore;
 import org.hypertrace.core.documentstore.Document;
 import org.hypertrace.core.documentstore.JSONDocument;
-import org.hypertrace.core.documentstore.Key;
 import org.hypertrace.core.documentstore.SingleValueKey;
 import org.hypertrace.core.grpcutils.client.GrpcChannelRegistry;
 import org.hypertrace.core.grpcutils.context.RequestContext;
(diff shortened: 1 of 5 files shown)


entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -289,128 +289,20 @@ private void doUpdate(RequestContext requestContext, EntityUpdateRequest request
...
-      responseObserver.onError(new ServiceException("Entity type is missing in the request."));
-      return;
-    }
-    if (request.getEntitiesCount() == 0) {
-      responseObserver.onError(new ServiceException("Entities are missing in the request."));
-    }
-    Map<String, EntityUpdateInfo> entitiesMap = request.getEntitiesMap();
-    try {
-      doBulkUpdate(requestContext, entitiesMap);
-      responseObserver.onCompleted();
```
aaron-steinfeld, line 340: will try to look through this more carefully at some point today, but one thing that jumps out at me off the bat - we never actually `onNext` a result set chunk here that I can see.

aaron-steinfeld (approved):

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
aaron-steinfeld: Discussed a bit offline. There may not be a bug here, but it has a very unintuitive api that can easily lead to corruption - a Value going in looks like a oneof style object but is not, so both the field and type need to be set. If they're not both set, we may be changing data types of persisted data.

skjindal93: Oh, the API isn't supposed to send any response, as of now. We can just change the API to just send an empty response as of now, instead of result set chunk

skjindal93:
No need to revert the APIs. The API implementation is fine as tested by the integration tests https://github.com/hypertrace/entity-service/pull/128. We need to enhance the entity query service to support updates for integer value types

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
aaron-steinfeld: If we're going to break the new api by changing its return type anyway, perhaps it makes sense to fix the value type there?

aaron-steinfeld:
> No need to revert the APIs. The API implementation is fine as tested by the integration tests #128. We need to enhance the entity query service to support updates for integer value types

it should support that - wasn't the issue on the caller side?

PR #128, opened 2021-09-02
chore(entity-query-service): better integration tests for update API

skjindal93:


avinashkolluru (approved):

entity-service/src/integrationTest/java/org/hypertrace/entity/service/service/EntityQueryServiceTest.java
```diff
@@ -734,4 +788,10 @@ private static Datastore getDatastore() {
 
     return attributesMap;
   }
+
+  private <T> void assertListEquals(List<T> expected, List<T> actual) {
```
aaron-steinfeld, line 792: why can't this just use assertEquals?

entity-service/src/integrationTest/java/org/hypertrace/entity/service/service/EntityQueryServiceTest.java
```diff
@@ -484,6 +489,11 @@ public void testBulkUpdate() {
             .setTenantId(TENANT_ID)
             .setEntityType(EntityType.API.name())
             .setEntityName("api1")
+            .putAttributes(
```
aaron-steinfeld, line 492: can we add an integer to the integration test, too?

entity-service/src/integrationTest/java/org/hypertrace/entity/service/service/EntityQueryServiceTest.java
skjindal93: I tried that. It wasn't working with the basic `assertEquals`, though it should have

Didn't spend much time on why it wasn't working

entity-service/src/integrationTest/java/org/hypertrace/entity/service/service/EntityQueryServiceTest.java
skjindal93: We don't support updating INT yet via EQS update APIs. Planning to add support for that in a separate PR

entity-service/src/integrationTest/java/org/hypertrace/entity/service/service/EntityQueryServiceTest.java
aaron-steinfeld: IMO, worth figuring out - something's off here. This check for example doesn't assert that their order matches, which is part of list equality.

entity-service/src/integrationTest/java/org/hypertrace/entity/service/service/EntityQueryServiceTest.java
aaron-steinfeld: we don't? maybe I'm misunderstanding or looking in the wrong spot, but it sure looks like we do - unless it's in the API but not the implementation?
```proto
message UpdateOperation {
  oneof operation {
    SetAttribute setAttribute = 1;
    // more update operations in the future
  }
}
```
->
```proto
message SetAttribute {
  ColumnIdentifier attribute = 1;
  LiteralConstant value = 2;
}
```
->
```proto
message LiteralConstant {
  Value value = 1;
}
```
->
```proto
message Value {
  ValueType valueType = 1;

  string string = 3;
  int64 long = 4;
  int32 int = 5;
  float float = 6;
  double double = 7;
  bytes bytes = 8;
  bool boolean = 9;
  sfixed64 timestamp = 15;
  repeated string string_array = 16;
  repeated int64 long_array = 17;
  repeated int32 int_array = 18;
  repeated float float_array = 19;
  repeated double double_array = 20;
  repeated bytes bytes_array = 21;
  repeated bool boolean_array = 22;
  map<string, string> string_map = 23;
}
```

aaron-steinfeld: oh, the implementation only has long in its switch, not int. That's really confusing, but is actually right (other than the fact that it should yell loudly and not have it in its interface)- entities only have long fields, not int.

So could we add long to the test?

PR #130, opened 2021-09-16
chore(entity-constants): api last called millis

skjindal93:


entity-service-api/src/main/proto/org/hypertrace/entity/constants/v1/entity_attribute.proto
```diff
@@ -34,6 +34,7 @@ enum ApiAttribute {
   API_ATTRIBUTE_DISCOVERY_FROM = 24 [(string_value) = "api_discovery_from"];
   API_ATTRIBUTE_DISCOVERY_STATE = 25 [(string_value) = "api_discovery_state"];
   API_ATTRIBUTE_DISCOVERY_TIMESTAMP = 26 [(string_value) = "api_discovery_timestamp"];
+  API_ATTRIBUTE_LAST_CALLED_MILLIS = 27 [(string_value) = "last_called_millis"];
```
aaron-steinfeld, line 37: I thought these were deprecated? Are we just adding a new attribute? If so, what's missing if we just add it as a plain attribute in the API scope?

aaron-steinfeld: So what I meant was that historically, we added three separate things:
1. the entity field name, added here
2. the mapping in the entity definition saying what attribute service attribute corresponds to that attribute
3. the attribute service attribute

More recently, we started just doing 3 - there's no real reason to give a separate name to the entity field (which forces having to map it). That direct mapping was implemented here - using the mapping if defined, else assuming the entity field name matches the attribute name: https://github.com/hypertrace/entity-service/blob/7305e9dab304837378fab9786a2ca0d18385905a/entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityAttributeMapping.java#L45-L55

Does that make sense? It doesn't quite answer the question of where to define the attribute's key for situations it is required in code, but it treats the entity fields the same as any other attribute (which we usually just define in place).

PR #131, opened 2021-09-23
Changes to publish entity change events

saxenakshitiz:


entity-service-api/src/main/proto/org/hypertrace/entity/change/event/v1/entity_change_event_key.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto3";
+
+option java_multiple_files = true;
+
+package org.hypertrace.entity.data.service.v1;
+
+message EntityChangeEventKey {
```
aaron-steinfeld, line 7: Can we make this a separate module (with its own package)? Regular API consumers need not get these protos.

entity-service-api/src/main/proto/org/hypertrace/entity/change/event/v1/entity_change_event_value.proto
```diff
@@ -0,0 +1,20 @@
...
+package org.hypertrace.entity.data.service.v1;
+
+import "org/hypertrace/entity/data/service/v1/entity.proto";
+
+message EntityChangeEventValue {
+  EventType event_type = 1;
+  optional Entity previous_version = 2; // previous will not be present for newly created entity
+  optional Entity latest_version = 3; // latest version will not be present for delete entity
+
+  enum EventType {
```
aaron-steinfeld, line 14: up to you - can probably replace the enum with a oneof containing three specific messages  (this just represents the comments you wrote directly in the API spec rather than in comments). We don't plan any future changes (famous last words) here, but it also would mean if we added another event type we need not make it match with these other fields.

entity-service/src/main/resources/configs/common/application.conf
```diff
@@ -78,3 +78,14 @@ metrics.reporter {
   console.reportInterval = 30
 }
 
+event.store {
```
aaron-steinfeld, line 81: We need to add something to the helm values to create the topic, right?

aaron-steinfeld (commented):
overall approach lgtm.

entity-service-change-event-api/build.gradle.kts
```diff
@@ -0,0 +1,68 @@
...
+        }
+      }
+    }
+  }
+}
+
+sourceSets {
+  main {
+    java {
+      srcDirs("src/main/java", "build/generated/source/proto/main/java", "build/generated/source/proto/main/grpc_java")
```
aaron-steinfeld, line 58: "src/main/java" is default, it's being duplicated here (that will lead to dupe protos which I think you mentioned seeing)

entity-service-change-event-api/src/main/java/org/hypertrace/entity/service/change/event/api/EntityChangeEventGenerator.java
```diff
@@ -0,0 +1,17 @@
...
+/** The interface Entity change event generator. */
+public interface EntityChangeEventGenerator {
+
+  /**
+   * Send change notification for any change to the entities
+   *
+   * @param existingEntityMap the map of existing entities
+   * @param updatedEntityMap the map of updated entities
+   */
+  void sendChangeNotification(
```
aaron-steinfeld, line 15: this seems like a strange interface except for update. Can we have a cleaner create/delete method? Now that I think about it, even on updates why not just take a list - if the generator needs to put that into a map, that's an impl detail. From the interface, it's not clear whaat the key is (I'm guessing entity ID?)

entity-service-change-event-api/build.gradle.kts
```diff
@@ -0,0 +1,68 @@
...
+    }
+  }
+}
+
+dependencies {
+  api("io.grpc:grpc-protobuf:1.40.1")
+  api("io.grpc:grpc-stub:1.40.1")
+  api("javax.annotation:javax.annotation-api:1.3.2")
+  protobuf(project(":entity-service-api"))
```
aaron-steinfeld, line 67: hmm - we really don't want proto packages depending on each other, which I guess is the problem wiith spliitting this out.

At a minimum, I think we'd change this to an API dependency, but that requires versioning it. Let's discuss this part.

entity-service/build.gradle.kts
```diff
@@ -14,13 +14,15 @@ tasks.integrationTest {
 
 dependencies {
   implementation(project(":entity-service-impl"))
+  implementation(project(":entity-service-change-event-impl"))
```
aaron-steinfeld, line 17: naming - it does use the new api but I think something like `:entity-service-change-event-generator` would be clearer.

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java
```diff
@@ -694,6 +716,25 @@ private JSONDocument convertEntityRelationshipToDocument(
     }
   }
 
+  private Map<String, Entity> searchExistingEntities(java.util.Collection<Entity> entities) {
+    return doQuery(buildExistingEntitiesInQuery(entities))
+        .collect(Collectors.toMap(entity -> getDocId(entity), Function.identity()));
```
aaron-steinfeld, line 721: why are we working with doc keys here?

entity-service-change-event-api/src/main/java/org/hypertrace/entity/service/change/event/api/EntityChangeEventGenerator.java
```diff
@@ -0,0 +1,17 @@
+package org.hypertrace.entity.service.change.event.api;
+
+import java.util.Map;
+import org.hypertrace.entity.data.service.v1.Entity;
+
+/** The interface Entity change event generator. */
+public interface EntityChangeEventGenerator {
```
aaron-steinfeld, line 7: I'd probably just put this in the same package as the impl. This isn't something we need published, but the protos we do. We can make the interface public and the impl package protected to keep the same separation we have here (just need a public builder or static build method somewhere).

entity-service-change-event-impl/build.gradle.kts
```diff
@@ -0,0 +1,38 @@
+plugins {
+  `java-library`
+  jacoco
+  id("org.hypertrace.jacoco-report-plugin")
+}
+
+dependencies {
+  api(project(":entity-service-api"))
```
aaron-steinfeld, line 8: very few of these deps are needed

entity-service-change-event-impl/build.gradle.kts
```diff
@@ -0,0 +1,38 @@
+plugins {
+  `java-library`
+  jacoco
```
aaron-steinfeld, line 3: please add tests (or remove jacoco - but please add tests 🙂  )

entity-service/src/integrationTest/resources/configs/entity-service/application.conf
```diff
@@ -58,6 +58,11 @@ entity.service.attributeMap = [
     "name": "API.apiDiscoveryState",
     "subDocPath": "attributes.api_discovery_state"
   },
+  {
+    "scope": "API",
```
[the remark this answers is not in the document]
saxenakshitiz: Removed

entity-service/build.gradle.kts
saxenakshitiz: done

entity-service-change-event-api/build.gradle.kts
saxenakshitiz: This is same as entity service protos. Doesn't fix the issue.

entity-service-change-event-impl/build.gradle.kts
saxenakshitiz: Done

entity-service-change-event-api/build.gradle.kts
surajpuvvada: just curious what is the problem if this depends on the other ? is it due to the possibility that somebody might import a certain version of entity-service-api and might have a different version of  entity-service-change-event-api ? I guess if both are on the same version then it shouldn't be a problem

entity-service-change-event-api/build.gradle.kts
```diff
@@ -0,0 +1,68 @@
...
+import com.google.protobuf.gradle.protobuf
+import com.google.protobuf.gradle.protoc
+
+plugins {
+  `java-library`
+  id("com.google.protobuf") version "0.8.15"
+  id("org.hypertrace.publish-plugin")
+}
+
+val generateLocalGoGrpcFiles = false
```
aaron-steinfeld, line 14: we don't need either the go or grpc targets here (if and when we need go, we actually handle that differently now anyway). A lot falls out - I think it resolves down to something like below
```kts

plugins {
  `java-library`
  id("com.google.protobuf") version "0.8.17"
  id("org.hypertrace.publish-plugin")
}

protobuf {
  protoc {
    artifact = "com.google.protobuf:protoc:3.18.0"
  }
}

dependencies {
  api("com.google.protobuf:protobuf-java:3.18.0")
}

sourceSets {
  main {
    java {
      srcDirs("build/generated/source/proto/main/java")
    }
  }
}
```

entity-service-change-event-api/build.gradle.kts
aaron-steinfeld: entity api file is wrong, at a minimum this is a redundant declaration, but this has been the issue in other places (due to that redundancy).

entity-service-change-event-generator/build.gradle.kts
```diff
@@ -0,0 +1,21 @@
+plugins {
+  `java-library`
+  jacoco
+  id("org.hypertrace.jacoco-report-plugin")
+}
+
+dependencies {
+  api(project(":entity-service-api"))
+  api(project(":entity-service-change-event-api"))
+  implementation("com.typesafe:config:1.4.1")
```
aaron-steinfeld, line 10: config is required to instantiate it, so should be part of public api rather than implementation.

entity-service-change-event-generator/src/main/java/org/hypertrace/entity/service/change/event/impl/EntityChangeEventGeneratorImpl.java
```diff
@@ -0,0 +1,123 @@
+package org.hypertrace.entity.service.change.event.impl;
+
+import static java.util.function.Function.identity;
+
+import com.google.common.collect.MapDifference;
```
aaron-steinfeld, line 5: Make sure to declare guava dep (likely getting it transitive as a grpc dependency right now, which we want to be explicit about)

entity-service-change-event-generator/src/main/java/org/hypertrace/entity/service/change/event/impl/EntityChangeEventGeneratorImpl.java
```diff
@@ -0,0 +1,123 @@
...
+    entityChangeEventProducer.send(getEntityChangeEventKey(currEntity), builder.build());
+  }
+
+  private void sendDeleteNotification(Entity deletedEntity) {
+    Builder builder = EntityChangeEventValue.newBuilder();
+    builder.setDeleteEvent(EntityDeleteEvent.newBuilder().setDeletedEntity(deletedEntity).build());
+    entityChangeEventProducer.send(getEntityChangeEventKey(deletedEntity), builder.build());
+  }
+
+  private EntityChangeEventKey getEntityChangeEventKey(Entity entity) {
```
aaron-steinfeld, line 116: discussed briefly last week, but may need to serialize something else for kafka keys given the protobuf non-determinism.

entity-service-impl/build.gradle.kts
```diff
@@ -15,6 +16,7 @@ dependencies {
   implementation("org.hypertrace.core.grpcutils:grpc-context-utils:0.6.1")
   implementation("org.hypertrace.core.grpcutils:grpc-client-utils:0.6.1")
   implementation("org.hypertrace.core.attribute.service:caching-attribute-service-client:0.12.3")
+  implementation("org.hypertrace.core.eventstore:event-store:0.1.2")
```
aaron-steinfeld, line 19: This is an impl of the generator project now.

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java
```diff
@@ -148,19 +165,26 @@ public void getAndUpsertEntities(Entities request, StreamObserver<Entity> respon
...
+      List<Entity> existingEntities =
+          Streams.stream(entitiesCollection.bulkUpsertAndReturnOlderDocuments(documentMap))
+              .flatMap(
+                  document -> PARSER.<Entity>parseOrLog(document, Entity.newBuilder()).stream())
+              .map(Entity::toBuilder)
+              .map(builder -> builder.setTenantId(tenantId))
+              .map(Builder::build)
+              .collect(Collectors.toList());
+
+      entityChangeEventGenerator.sendChangeNotification(existingEntities, updatedEntities);
```
aaron-steinfeld, line 186: We need to be careful, especially on bulk upsert, about error handling. If this fails to upsert something for example, does it return it? Does it return if no change? etc. We don't want to accidentally send a delete notification if no delete actually happened.

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java
```diff
@@ -148,19 +165,26 @@ public void getAndUpsertEntities(Entities request, StreamObserver<Entity> respon
     try {
 
       Map<Key, Document> documentMap = new HashMap<>();
+      List<Entity> updatedEntities = new ArrayList<>();
       for (Entity entity : request.getEntityList()) {
         Entity normalizedEntity = this.entityNormalizer.normalize(tenantId, entity);
+        updatedEntities.add(normalizedEntity);
```
aaron-steinfeld, line 171: Aren't these the entities we're about to upsert? So they'd be the new state rather than the old (would also include entities that don't yet exist and are about to be created).

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java
```diff
@@ -108,6 +121,8 @@ public void upsert(Entity request, StreamObserver<Entity> responseObserver) {
           Entity.newBuilder(),
           entitiesCollection,
           responseObserver);
+      entityChangeEventGenerator.sendChangeNotification(
```
aaron-steinfeld, line 124: in terms of error handling in general, we probably should not fail the call if this publish fails (that is, it should be outside the try).

entity-service-change-event-generator/src/main/java/org/hypertrace/entity/service/change/event/impl/EntityChangeEventGeneratorImpl.java
surajpuvvada: A string concatenated can be used

entity-service-change-event-api/build.gradle.kts
saxenakshitiz: Removed

kotharironak:
@saxenakshitiz Can we have a github context ticket for upcoming changes for moving from sync to async entity updates in the ingestion pipeline? we can log the issue here - https://github.com/hypertrace/hypertrace/issues

entity-service-change-event-api/build.gradle.kts
saxenakshitiz: done

entity-service-change-event-generator/src/test/java/org/hypertrace/entity/service/change/event/impl/EntityChangeEventGeneratorImplTest.java
```diff
@@ -0,0 +1,156 @@
...
+
+  @BeforeEach
+  void setup() {
+    changeEventGenerator = new EntityChangeEventGeneratorImpl(eventStore, eventProducer);
+  }
+
+  @Test
+  void sendCreateNotification() {
+    List<Entity> entities = createEntities(2);
+    ArgumentCaptor<EntityChangeEventKey> keyCaptor =
```
aaron-steinfeld, line 46: why use captors rather than plain matchers?

helm/values.yaml
```diff
@@ -113,3 +113,21 @@ hpa:
...
+  helmHook: pre-install,pre-upgrade
+  kafka:
+    topics:
+      entity-change-events:
+        replicationFactor: 3
+        partitions: 8
+        configs:
+          retention.bytes: 8589934592 # default = -1
+          retention.ms: 86400000 # default = 604800000 (7 days)
+          max.message.bytes: 10485760 # Allow larger messages for traces
```
aaron-steinfeld, line 129: Are these values correct for a low volume topic like this? Entities are much smaller than traces, and far lower volume so seems like the partitions and maxs could be reduced.

entity-service-change-event-generator/src/main/java/org/hypertrace/entity/service/change/event/impl/EntityChangeEventGeneratorImpl.java
aaron-steinfeld: this seems like an important one to follow up on one way or the other - we don't want to change the key later.

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java
```diff
@@ -108,6 +121,7 @@ public void upsert(Entity request, StreamObserver<Entity> responseObserver) {
           Entity.newBuilder(),
           entitiesCollection,
           responseObserver);
+      sendChangeNotification(tenantId, existingEntityCollection, List.of(normalizedEntity));
```
aaron-steinfeld, line 124: should this move out of the try? we should never onError after we've already onComplete'd which I'm guessing is happening in upsertEntity. Same on all others. Even if we hadn't completed already, the rpc calls to ES should never fail due to the notifications of the changes. IMO, we should swallow and log all errors in the publisher and always call it as the last line of the success branch, after onComplete.

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java
```diff
@@ -148,19 +164,26 @@ public void getAndUpsertEntities(Entities request, StreamObserver<Entity> respon
...
+      List<Entity> existingEntities =
+          Streams.stream(entitiesCollection.bulkUpsertAndReturnOlderDocuments(documentMap))
+              .flatMap(
+                  document -> PARSER.<Entity>parseOrLog(document, Entity.newBuilder()).stream())
+              .map(Entity::toBuilder)
+              .map(builder -> builder.setTenantId(tenantId))
+              .map(Builder::build)
+              .collect(Collectors.toList());
+
+      sendChangeNotification(tenantId, existingEntities, updatedEntities);
```
aaron-steinfeld, line 185: most of our notifications happen after we notify the response observer, this one happens before. We should be consistent (after, IMO).

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java
```diff
@@ -539,10 +562,13 @@ public void mergeAndUpsertEntity(
               .map(Builder::build)
               .orElse(receivedEntity);
       try {
+        Entity upsertedEntity = this.upsertEntity(tenantId, entityToUpsert);
+        sendChangeNotification(
+            tenantId,
+            existingEntity.map(e -> List.of(e)).orElse(Collections.emptyList()),
```
aaron-steinfeld, line 568: nit: List::of

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java
```diff
@@ -694,6 +720,58 @@ private JSONDocument convertEntityRelationshipToDocument(
     }
   }
 
+  private void sentDeleteNotification(String tenantId, java.util.Collection<Entity> entities) {
```
aaron-steinfeld, line 723: oh, so you've already done most of what I was talking about in the other comment - we swallow errors. Since the generator is local to this service and this is its only caller, we can push these down inside.

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java
```diff
@@ -694,6 +720,58 @@ private JSONDocument convertEntityRelationshipToDocument(
...
+    try {
+      entityChangeEventGenerator.sendDeleteNotification(entities);
+    } catch (Exception ex) {
+      LOG.warn(
+          "For tenantId {] unable to send delete notification for entities {}",
```
aaron-steinfeld, line 728: typo on the placeholder, but I'd encourage dropping the tenantID arg here. It's already part of the entity and thus is duplicate (and even if it weren't we could look it up if needed).

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java
```diff
@@ -694,6 +720,58 @@ private JSONDocument convertEntityRelationshipToDocument(
...
+          getEntityIds(existingEntities),
+          getEntityIds(updatedEntities));
+    }
+  }
+
+  private List<String> getEntityIds(java.util.Collection<Entity> entities) {
+    return entities.stream().map(Entity::getEntityId).collect(Collectors.toList());
+  }
+
+  private java.util.Collection<Entity> getExistingEntities(java.util.Collection<Entity> entities) {
```
aaron-steinfeld, line 753: nit: makes no real difference here but... if we have a list, might as well return a list - it can still be used as a collection (accept liberally, return conservatively)

saxenakshitiz:
These changes relate to https://github.com/hypertrace/hypertrace/issues/306

entity-service-change-event-generator/src/test/java/org/hypertrace/entity/service/change/event/impl/EntityChangeEventGeneratorImplTest.java
aaron-steinfeld: Maybe I'm forgetting my mockito, I thought you could just as easily do something like below, which is more readable IMO
```java
verify(eventProducer, times(1)).send(eq(expectedFirstKey), eq(expectedFirstValue));
verify(eventProducer, times(1)).send(eq(expectedSecondKey), eq(expectedSecondValue));
```

entity-service-change-event-api/src/main/java/org/hypertrace/entity/service/change/event/api/EntityChangeEventGenerator.java
aaron-steinfeld: Nah, that's fine.

entity-service-change-event-generator/src/main/java/org/hypertrace/entity/service/change/event/impl/EntityChangeEventGeneratorImpl.java
aaron-steinfeld: > Entity id is random enough to distribute well across partition. Also entity type and tenant id are available as part of entity itself.

This depends on the entity type. If there's no penalty (the consumer doesn't care what the key is - right?), I'd go for the concatenated key. For V1 entities, the entity ID is hashed across identifying attributes, type and tenant id so should be globally unique and fine, but for V2 entities, the entity ID is opaque to entity service and determined by the definition. The uniqueness across types and tenant IDs for v2 entities is treated as an internal impl detail and is only accounted for at the doc key. What this means is, for example, imagine a v2 service entity (service is currently v1, but it explains the general issue for v2 entities) which is identified by the servicename attribute. Across tenants, the name "data-service" or "iam" for example, may be common resulting in identical user-facing entityIds (but unique doc ids).

Is it going to be random enough if we just use the entity id? I've got to imagine yes, and I'm fine with that if there's any real downside. It just seems like cheap insurance to concatenate the other information, too (for the record, that's exactly how the doc key is generated for v2 entities, too - `String.format("%s:%s:%s", this.tenantId, this.entityType, this.entityId)`).

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java
aaron-steinfeld: I think I did another round of review on the newer code before this set of comments :)

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java
aaron-steinfeld: I'm just not sure what the error expectation is on bulk upsert in terms of atomicity and consistency. If it upserts 9 and then the 10th fails, does the whole call fail (and if so, does it roll the first 9 back?). Any changes there are out of scope, I just want to make sure we match the behavior because it's implementation-dependent.

One mismatch I can definitely see: parseOrLog. If I upsert 3 entities (all previously existing), and get the 3 originals back, but we fail to parse one, parseOrLog will log and drop it so as to not fail the whole call - whatever was malformed was overwritten, anyway. We'd end up with an array of 2 existing entities and 3 updated entities, which as I understand would send out a create event incorrectly. As to what we'd expect to happen there... I guess an update event and an unknown prior state?

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java
aaron-steinfeld: Ideally, a rename too would help to convey to the caller that this won't fail. Like `trySendDeleteNotification`.

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java
saxenakshitiz: done

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java
saxenakshitiz: done

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java
saxenakshitiz: Renamed. Just cleaner to have a separate method

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java
saxenakshitiz: Fixed

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java
saxenakshitiz: Done

helm/values.yaml
saxenakshitiz: Fixed after discussing with @laxmanchekka and @ravisingal

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java
saxenakshitiz: As you have reviewed, this part of fixed.

entity-service-change-event-generator/src/test/java/org/hypertrace/entity/service/change/event/impl/EntityChangeEventGeneratorImplTest.java
saxenakshitiz: This will work fine as well. ArgumentCaptor allows to validate order as well.

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java
saxenakshitiz: For bulkupsert, I am assuming all or none. Can validate that part.

parseOrLog is interesting. I can pull existing entities first and not depend on parseOrLog at all. That may be better.

entity-service-change-event-generator/src/test/java/org/hypertrace/entity/service/change/event/impl/EntityChangeEventGeneratorImplTest.java
aaron-steinfeld: verify supports that too - https://github.com/hypertrace/java-grpc-utils/blob/cb552a00c4967fee19000386f662bfaec7e11f1c/grpc-client-utils/src/test/java/org/hypertrace/core/grpcutils/client/GrpcChannelRegistryTest.java#L96-L103 for example.


Anyway, this was just a nit for readability.

entity-service-change-event-generator/src/main/java/org/hypertrace/entity/service/change/event/impl/EntityChangeEventGeneratorImpl.java
saxenakshitiz: Used logic similar to doc key to generate the key

entity-service-change-event-generator/src/test/java/org/hypertrace/entity/service/change/event/impl/EntityChangeEventGeneratorImplTest.java
saxenakshitiz: Anyways removed capturs. In some cases, it was adding just additional complexity.

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java
```diff
@@ -720,18 +721,18 @@ private JSONDocument convertEntityRelationshipToDocument(
     }
   }
 
-  private void sentDeleteNotification(String tenantId, java.util.Collection<Entity> entities) {
+  private void trySendDeleteNotification(String tenantId, java.util.Collection<Entity> entities) {
```
aaron-steinfeld, line 724: my last remaining comment on this, and its up to you, is that as the only consumer of `entityChangeEventGenerator`, we shouldn't have to wrap its methods. I'd push this try/catch/log into the generator itself.

aaron-steinfeld (approved):

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java
saxenakshitiz: Done

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java
saxenakshitiz: we use bulkwrite API on mongo side - https://docs.mongodb.com/manual/reference/method/db.collection.bulkWrite

This allows for partial writes. So we need to pull entities on task completion and use that. Correct?

aaron-steinfeld: so it's kind of confusing. If it does do a partial write, let's go through a scenario.

1. Imagine I start with X, Y persisted.
2. I upsert X', Y', Z.
3. Y' fails to upsert for whatever reason. I'm not sure how this is reflected in the bulk upsert result, if at all (is Y omitted from the previous entities returned?)
   -  In the current impl, we don't use the upsert results, so would include an update of Y-> Y' that never happened (in addition to a correct creation of Z and update of X->X')
   - In the previous impl, we'd be wrong in one of two different ways, depending on what comes back in the results:
      - If Y is omitted from the previous entity result, than we'd be showing a Y' creation when Y already existed _and_ Y' failed to update
      - If Y is not omitted from previous entity result, we'd be showing a Y->Y' update that never happened. 

So this is tricky, and it does depend on the behavior of bulkwrite, but my suspicion is our bulk upsert API does not give us the info we need to completely solve this. If it omits the failed update from the previous state, we can't distinguish a failed update from a create. If it includes it, we just don't know if it succeeded or not.

Rather than make any changes in the response, I think I would want us to prevent any partial updates from happening (or verify that this is already the behavior) in doc store - all or nothing just has far less ambiguity. As long as we can trust that the updates all happened on success, then we can go back to what we had - we don't need the extra fetch and can use the prior state result of the bulk upsert (we just want to make sure that any deserialization errors of the prior state don't lead to erroneous creation notifications).

aaron-steinfeld: > we just want to make sure that any deserialization errors of the prior state don't lead to erroneous creation notifications).

I just looked back and this is pretty widespread, it may be tricky to truly account for. Anywhere we call `doQuery`, underneath we do this log and drop behavior. I guess being consistent is better unless you want to try to tackle that

entity-service/build.gradle.kts
```diff
@@ -22,7 +22,7 @@ dependencies {
   implementation("org.hypertrace.core.documentstore:document-store:0.6.0")
 
   runtimeOnly("io.grpc:grpc-netty:1.40.1")
-  runtimeOnly("io.confluent:kafka-protobuf-serializer:6.0.1")
+  runtimeOnly("io.confluent:kafka-protobuf-serializer:6.2.1")
```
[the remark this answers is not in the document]
saxenakshitiz: Ideally should not. Anyways I will deploy these changes first on dave cluster to check.

entity-service-change-event-generator/src/main/java/org/hypertrace/entity/service/change/event/impl/EntityChangeEventGeneratorFactory.java
```diff
@@ -0,0 +1,27 @@
...
+
+  static final String ENTITY_SERVICE_PUBLISH_CHANGE_EVENTS =
+      "entity.service.config.publish.change.events";
+
+  private static final EntityChangeEventGeneratorFactory instance =
+      new EntityChangeEventGeneratorFactory();
+
+  private EntityChangeEventGeneratorFactory() {}
+
+  public static EntityChangeEventGeneratorFactory getInstance() {
```
aaron-steinfeld, line 16: Does a static `getInstance()` give us any benefit over a static method `createEntityChangeEventGenerator` ? Or just a style difference?

saxenakshitiz: Just a style difference

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java
```diff
@@ -130,7 +145,9 @@ public void upsertEntities(Entities request, StreamObserver<Empty> responseObser
                   Collectors.toUnmodifiableMap(
                       entity -> this.entityNormalizer.getEntityDocKey(tenantId, entity),
                       Function.identity()));
+      java.util.Collection<Entity> existingEntities = getExistingEntities(entities.values());
```
aaron-steinfeld, line 148: So if we assume doc store behaves consistently (discussed elsewhere), then we don't need this extra query, the result of the original docs returned on upsert should be sufficient.

saxenakshitiz: upsertEntities() doesn't return old documents.  Probably you meant `getAndUpsertEntities` method which was calling `bulkUpsertAndReturnOlderDocuments`. I have already reverted my change there. Please have a look.

aaron-steinfeld: you're right, I was, carry on!

aaron-steinfeld (approved):

entity-service/src/main/resources/configs/common/application.conf
```diff
@@ -78,3 +79,14 @@ metrics.reporter {
...
+  type = kafka
+  bootstrap.servers = "localhost:9092"
+  entity.change.events.producer {
+    topic.name = entity-change-events
+    bootstrap.servers = "localhost:9092"
+    key.serializer = org.apache.kafka.common.serialization.StringSerializer
+    value.serializer = io.confluent.kafka.serializers.protobuf.KafkaProtobufSerializer
+    schema.registry.url = "http://localhost:8081"
```
avinashkolluru, line 90: A little late to the review but I'm assuming `schema.registry.url` is not needed. The schema registry is specific to avro messages but here its proto.

saxenakshitiz: Actually schema registry also supports protos. We are using version confluent version 5.5 which does support protos schema.

entity-service-change-event-generator/src/main/java/org/hypertrace/entity/service/change/event/util/KeyUtil.java
```diff
@@ -0,0 +1,13 @@
...
+
+import org.hypertrace.entity.data.service.v1.Entity;
+
+public class KeyUtil {
+
+  private KeyUtil() {}
+
+  public static final String getKey(Entity entity) {
+    return String.format(
+        "%s:%s:%s", entity.getTenantId(), entity.getEntityType(), entity.getEntityId());
```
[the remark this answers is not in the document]
saxenakshitiz: All this information is already available within Entity object, so key only purpose right now is to send entity changes to same partition.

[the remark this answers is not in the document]
aaron-steinfeld: Right, we actually revisited this conversation yesterday. Conclusion was basically what both of you are saying:
If consumers need to understand the key, it should be a proto message. If the key is purely for partitioning, we can treat it as opaque and send whatever format that gives us that fair distribution.

PR #132, opened 2021-10-01
fix(eqs): check value type set for eqs update

skjindal93:


skjindal93:
fixing the unit tests

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryConverter.java
```diff
@@ -337,10 +343,22 @@ private AttributeFilter convertToAttributeFilter(
...
+      case UNRECOGNIZED:
+      default:
+        throw new IllegalArgumentException(
+            String.format("Unrecognized value type %s for literal %s", valueType, literal));
     }
     return builder;
   }
 
+  private static boolean isValueTypeSet(org.hypertrace.entity.query.service.v1.Value value) {
+    return value.hasField(
```
aaron-steinfeld, line 355: Are we trying to distinguish between unset and set to default? I'm fairly sure this won't work and almost positive we can't _guarantee_ it will work. Proto is free to represent a value in whatever way it can for transport, we can't assume it would represent set to default and unset any differently.

aaron-steinfeld (changes requested):

entity-service-impl/src/test/java/org/hypertrace/entity/query/service/EntityQueryConverterTest.java
```diff
@@ -193,6 +195,33 @@ public void test_convertEqsSelections_DocStoreSelections() {
...
+            .build();
+
+    assertThrows(
+        IllegalArgumentException.class,
+        () -> EntityQueryConverter.convertToAttributeValue(literal));
+  }
+
+  @Test
+  void testValidAttributeConversion() {
+    LiteralConstant literal =
```
aaron-steinfeld, line 212: For example, if I change this to the more interesting test, explicitly setting a string, we can see it fail:

```diff
diff --git a/entity-service-impl/src/test/java/org/hypertrace/entity/query/service/EntityQueryConverterTest.java b/entity-service-impl/src/test/java/org/hypertrace/entity/query/service/EntityQueryConverterTest.java
index 642dfa8..062e546 100644
--- a/entity-service-impl/src/test/java/org/hypertrace/entity/query/service/EntityQueryConverterTest.java
+++ b/entity-service-impl/src/test/java/org/hypertrace/entity/query/service/EntityQueryConverterTest.java
@@ -213,12 +213,12 @@ public class EntityQueryConverterTest {
         LiteralConstant.newBuilder()
             .setValue(
                 org.hypertrace.entity.query.service.v1.Value.newBuilder()
-                    .setLong(123L)
-                    .setValueType(ValueType.LONG))
+                    .setString("test")
+                    .setValueType(ValueType.STRING))
             .build();
 
     assertEquals(
-        AttributeValue.newBuilder().setValue(Value.newBuilder().setLong(123L).build()).build(),
+        AttributeValue.newBuilder().setValue(Value.newBuilder().setString("test").build()).build(),
         EntityQueryConverter.convertToAttributeValue(literal).build());
   }
```
it throws:
```
java.lang.IllegalArgumentException: Value type not set for literal value {
  string: "test"
}
```

skjindal93: I see. I don't think we can add a validation for the case, when the value type isn't set on the Value

aaron-steinfeld: Yeah don't think it's possible with the current proto.

skjindal93: Wondering would it be worth while making a breaking change by adding optional for every field 


```
message Value {
  ValueType valueType = 1;

  optional string string = 3;
  optional int64 long = 4;
  optional int32 int = 5;
  optional float float = 6;
  ......
}
```

And, in the code, during the value type switch case, we can check whether the optional is set for that particular type. 

Since, we are adding optionals, the consumers of the API won't break, but I believe it would just look bad in terms of the API contract

aaron-steinfeld: If we're going to make a breaking change, might as well fix it right.

skjindal93: Is there any other way apart from extending the API via `oneof` for setting the values and deprecating the current way of setting value with value type?

aaron-steinfeld: cleanest way to do it fully compatible I think would be to deprecate the whole value message, the field on literal constant (And anywhere else its used - not sure if it's widespread?), and add a new field referring to a correct version of value.

Probably not worth the effort right now.

PR #134, opened 2021-10-06
Enhanced EdsCacheClient to support async reloading and refreshAfterWrite

avinashkolluru:
## Description
Enhanced the EdsCacheClient with support for configuring refreshAfterWrite and also async cache reloading.

### Checklist:

### DIFF
--- a/entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
+++ b/entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
@@ -9,6 +9,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
 import javax.annotation.Nonnull;
 import org.hypertrace.core.serviceframework.metrics.PlatformMetricsRegistry;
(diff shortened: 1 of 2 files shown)


entity-service-client/src/main/java/org/hypertrace/entity/service/client/config/EntityServiceClientCacheConfig.java
```diff
@@ -10,28 +10,39 @@
  */
 public class EntityServiceClientCacheConfig {
 
+  public static long DEFAULT_CACHE_REFRESH_MS = 240000L;
```
kotharironak, line 13: Will this enable expiry to extend by another 5 mins? Does this mean that we don't want to expire the entry?

kotharironak: So, if the replacement has happened, it will continue to stay in memory, But, if a replacement has not happened, it will be expired, right?

avinashkolluru: > So, if the replacement has happened, it will continue to stay in memory, But, if a replacement has not happened, it will be expired, right?
Yes but a small correction. expiry in guava cache doesn't mean the entry will be cleaned up from the cache automatically. There is no auto cleanup of expired entries. Its just that once expired, the stale entry in the cache is not returned and rather the loader is invoked to get the fresh value. Entries are evicted from the cache only based on the size.

kotharironak: Got it.

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
```diff
@@ -27,75 +28,95 @@
...
+                          throw new NotFoundException("Entity not found!!");
+                        }
+                        entityCache.put(
+                            new EdsCacheKey(entity.getTenantId(), entity.getEntityId()), entity);
+                        return entity.getEntityId();
+                      }
+                    },
+                    executor));
+    PlatformMetricsRegistry.registerCache(
+        this.getClass().getName() + ".enrichedEntityCache",
```
kotharironak, line 113: Curious: Any reason for prefixing with the class name?

avinashkolluru: easier to track which class holds the caches and also avoid naming conflicts

PR #135, opened 2021-10-12
Modified to have placeholder for empty values and maps

saxenakshitiz:


entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -403,7 +403,8 @@ private void doBulkUpdate(
       AttributeValue attributeValue =
           EntityQueryConverter.convertToAttributeValue(setAttribute.getValue()).build();
       try {
-        String jsonValue = DocStoreJsonFormat.printer().print(attributeValue);
+        String jsonValue =
+            DocStoreJsonFormat.printer().includingDefaultValueFields().print(attributeValue);
```
aaron-steinfeld, line 407: Can we save a static for the printer like we do the parser so we can use consistent settings?

Also, besides the larger doc size, can you think of any implications? My biggest concern is that this might change the result of `.hasXyz()` for unset fields.

saxenakshitiz: Good to merge this change then?

PR #136, opened 2021-10-12
Added proto key for entity change events

saxenakshitiz:
1. Added proto key for entity change events
2. Added user details to entity change event value


surajpuvvada (approved):
lgtm. 
Just wondering if we should only pass the fields that are needed from RequestContext or pass the entire request context ?

aaron-steinfeld:
> Keep it more flexible in case we want to expose any other information from request context in future. Changing method signature will require too many changes without much benefit.

+1 - we almost always want to pass the whole request context. Pulling info out of it should only be done at the leaf (because if we make a downstream call, we should be propagating the whole context), so asking for something like tenant id or another field off request context on the API breaks abstraction.

aaron-steinfeld (approved):

PR #137, opened 2021-10-12
Update pr-build.yml

saxenakshitiz:


aaron-steinfeld (dismissed):
Believe this is failing because the version field didn't exist on 0.18.1 and the workflow changes don't take hold til after merge. I'm fine either overriding or if you prefer, we can move the buf.yaml changes into the following PR.

PR #142, opened 2021-10-21
Changes to perform bulk operation on attribute with array value

saxenakshitiz:
1. Exposing it in EntityQueryService as other required helper classes and method exists in that class. Don't like it. However exposing it in EntityDataService will require some major code refactoring.
2. Hard-coded suffix for array value to ".valueList.values".  Suggestions are welcome.

### DIFF
--- a/entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/entity_query_request.proto
+++ b/entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/entity_query_request.proto
@@ -50,3 +50,18 @@ message SetAttribute {
   ColumnIdentifier attribute = 1;
   LiteralConstant value = 2;
 }
+
+message BulkEntityArrayAttributeUpdateRequest {
+  string entityType = 1;
+  repeated string entityIds = 2;
+  ColumnIdentifier attribute = 3;
+  Operation operation = 4;
+  repeated LiteralConstant values = 5;
+
+  enum Operation {
+    OPERATION_UNSPECIFIED = 0;
+    OPERATION_ADD = 1;
+    OPERATION_REMOVE = 2;
+    OPERATION_SET = 3;
+  }
+}
--- a/entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/entity_query_request.proto
+++ b/entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/entity_query_request.proto
@@ -52,8 +52,8 @@ message SetAttribute {
 }
 
 message BulkEntityArrayAttributeUpdateRequest {
-  string entityType = 1;
-  repeated string entityIds = 2;
+  string entity_type = 1;
+  repeated string entity_ids = 2;
   ColumnIdentifier attribute = 3;
   Operation operation = 4;
   repeated LiteralConstant values = 5;
@@ -64,4 +64,7 @@ message BulkEntityArrayAttributeUpdateRequest {
     OPERATION_REMOVE = 2;
     OPERATION_SET = 3;
   }
+}
+
+message BulkEntityArrayAttributeUpdateResponse {
 }
--- a/entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/entity_query_service.proto
+++ b/entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/entity_query_service.proto
@@ -18,6 +18,6 @@ service EntityQueryService {
   }
   rpc total (TotalEntitiesRequest) returns (TotalEntitiesResponse) {
   }
-  rpc bulkUpdateEntityArrayAttribute (BulkEntityArrayAttributeUpdateRequest) returns (stream ResultSetChunk) {
+  rpc bulkUpdateEntityArrayAttribute (BulkEntityArrayAttributeUpdateRequest) returns (BulkEntityArrayAttributeUpdateResponse) {
   }
 }
--- a/entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/entity_query_service.proto
+++ b/entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/entity_query_service.proto
@@ -18,4 +18,6 @@ service EntityQueryService {
   }
   rpc total (TotalEntitiesRequest) returns (TotalEntitiesResponse) {
   }
+  rpc bulkUpdateEntityArrayAttribute (BulkEntityArrayAttributeUpdateRequest) returns (stream ResultSetChunk) {
+  }
 }
--- a/entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
+++ b/entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
@@ -1,6 +1,9 @@
 package org.hypertrace.entity.query.service;
 
 import static java.util.Objects.isNull;
+import static java.util.stream.Collectors.joining;
+import static org.hypertrace.entity.data.service.v1.AttributeValue.VALUE_LIST_FIELD_NUMBER;
+import static org.hypertrace.entity.data.service.v1.AttributeValueList.VALUES_FIELD_NUMBER;
 import static org.hypertrace.entity.query.service.EntityAttributeMapping.ENTITY_ATTRIBUTE_DOC_PREFIX;
 import static org.hypertrace.entity.service.constants.EntityCollectionConstants.RAW_ENTITIES_COLLECTION;
 
@@ -18,6 +21,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import lombok.SneakyThrows;
 import org.hypertrace.core.documentstore.BulkArrayValueUpdateRequest;
 import org.hypertrace.core.documentstore.Collection;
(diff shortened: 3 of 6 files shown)


entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/entity_query_request.proto
```diff
@@ -50,3 +50,18 @@ message SetAttribute {
   ColumnIdentifier attribute = 1;
   LiteralConstant value = 2;
 }
+
+message BulkEntityArrayAttributeUpdateRequest {
+  string entityType = 1;
```
aaron-steinfeld, line 55: Match buf conventions please (guessing it's not on in this repo, but we should enforce for new stuff). Here at least, camelCase vs snake_case.

entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/entity_query_request.proto
```diff
@@ -50,3 +50,18 @@ message SetAttribute {
...
+  repeated string entityIds = 2;
+  ColumnIdentifier attribute = 3;
```
aaron-steinfeld, line 57: it may make sense to pull the operation into its own message and take multiple so we can make multiple updates at once. Would the lower level api support that? If not, I guess we can defer that and roll it into my dream of a more complex bulk update api.

entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/entity_query_service.proto
```diff
@@ -18,4 +18,6 @@ service EntityQueryService {
   }
   rpc total (TotalEntitiesRequest) returns (TotalEntitiesResponse) {
   }
+  rpc bulkUpdateEntityArrayAttribute (BulkEntityArrayAttributeUpdateRequest) returns (stream ResultSetChunk) {
```
aaron-steinfeld, line 21: we don't need to (or want to) stream results. We've stopped doing that for new rpcs - can we give it a plain unary response like `BulkEntityArrayAttributeUpdateResponse`

entity-service/build.gradle.kts
```diff
@@ -19,7 +19,7 @@ dependencies {
   implementation("org.hypertrace.core.grpcutils:grpc-server-utils:0.6.1")
   implementation("org.hypertrace.core.grpcutils:grpc-client-utils:0.6.1")
   implementation("org.hypertrace.core.serviceframework:platform-service-framework:0.1.28")
-  implementation("org.hypertrace.core.documentstore:document-store:0.6.0")
+  implementation("org.hypertrace.core.documentstore:document-store:0.6.1")
```
aaron-steinfeld, line 22: Does the service container use doc store directly? If yes, change is good as is. If no, we can remove the dependency entirely.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -346,6 +350,70 @@ public void bulkUpdate(
...
+    String tenantId = requestContext.getTenantId().orElse(null);
+    if (isNull(tenantId)) {
+      responseObserver.onError(new ServiceException("Tenant id is missing in the request."));
+      return;
+    }
+    try {
+      Set<Key> keys =
+          request.getEntityIdsList().stream()
+              .map(entityId -> new SingleValueKey(tenantId, entityId))
+              .collect(Collectors.toSet());
```
aaron-steinfeld, line 367: collect to list (or ordered set - point is to maintain order)

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -346,6 +350,70 @@ public void bulkUpdate(
...
+
+      String attributeId = request.getAttribute().getColumnName();
+
+      String subDocPath =
+          entityAttributeMapping
+              .getDocStorePathByAttributeId(requestContext, attributeId)
+              .orElseThrow(
+                  () -> new IllegalArgumentException("Unknown attribute FQN " + attributeId));
```
aaron-steinfeld, line 375: nit (FQN in the context of attributes is deprecated, we use attribute IDs now):
```suggestion
                  () -> new IllegalArgumentException("Unknown attribute " + attributeId));
```

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -346,6 +350,70 @@ public void bulkUpdate(
...
+
+      List<Document> subDocuments = new ArrayList<>();
```
aaron-steinfeld, line 377: nit: stream/map/collect to unmodifiable

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -346,6 +350,70 @@ public void bulkUpdate(
...
+      for (LiteralConstant literalConstant : request.getValuesList()) {
+        subDocuments.add(convertToJsonDocument(literalConstant));
+      }
+      BulkArrayValueUpdateRequest bulkArrayValueUpdateRequest =
+          new BulkArrayValueUpdateRequest(
+              keys,
+              subDocPath + ARRAY_VALUE_PATH_SUFFIX,
```
aaron-steinfeld, line 384: How do we handle this elsewhere? Do we write the whole value doc and here we're only doing a subset of the value doc? I wonder if there's a way for us to pull this out of the proto, since this is based on the shape of the `AttributeValueList` message. That said, aren't the subdocuments we're printing here `AttributeValue`s? That seems like a mismatch to me. I'd expect this path to show up in the subdocs too (which would prevent them from merging).

Can we add a test that verifies the `BulkArrayValueUpdateRequest` ?

entity-service/src/integrationTest/java/org/hypertrace/entity/service/service/EntityQueryServiceTest.java
```diff
@@ -639,6 +641,74 @@ public void testBulkUpdateWithLabels() {
...
+                        .build(),
+                    LiteralConstant.newBuilder()
+                        .setValue(
+                            org.hypertrace.entity.query.service.v1.Value.newBuilder()
+                                .setString("Label4")
+                                .build())
+                        .build()))
+            .build();
+
+    assertDoesNotThrow(
```
aaron-steinfeld, line 706: We should verify the results. Did the labels actually merge?

entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/entity_query_request.proto
saxenakshitiz: Done

entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/entity_query_request.proto
saxenakshitiz: Not in its current form.

Also the frontend APIs may be scoped smaller and one at the backend(data store one) may handle multiple of these smaller bulk operations

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
saxenakshitiz: ordering for keys matter only if we want to update them in a specific order. I don't think that is the case here.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
saxenakshitiz: I tried to extract it from AttributeValue -> AttributeValueList. It was just too complicated to extract that path.

AttributeValueList actually contains repeated AttributeValue. 
```
message AttributeValueList {
  repeated AttributeValue values = 1;
}
```

entity-service/src/integrationTest/java/org/hypertrace/entity/service/service/EntityQueryServiceTest.java
saxenakshitiz: Added

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
aaron-steinfeld: Ok, got it - my mistake. My confusion was that I thought for some reason we were converting an `AttributeValue` holding a list, rather than a list of individual `AttributeValue` scalars. The full path would look like
`attributes.<attrName>.valueList.values[value.string]` for a list of strings, and we're passing an array of docs for path `attributes.<attrName>.valueList.values`, those docs each representing a `value.string` so all good.


As for extracting the path, it's a bit more verbose, but shouldn't be particularly complicated:
```java
Stream.of(
        AttributeValue.getDescriptor().findFieldByNumber(VALUE_LIST_FIELD_NUMBER).getJsonName(),
        AttributeValueList.getDescriptor().findFieldByNumber(VALUES_FIELD_NUMBER).getJsonName()
).collect(joining("."))
```

PR #148, opened 2021-12-06
fix: map api service name

aaron-steinfeld:
## Description
This is the second half of fix for https://github.com/hypertrace/hypertrace-ui/issues/1061 . In that ticket, previously @skjindal93 started populating serviceName on API entities but instead of using the new attributeId which would be mapped automatically, it used the old entity service key which requires an explicit subpath mapping. This PR adds that mapping.

### Testing
Built and ran locally, verified behavior.

Closes https://github.com/hypertrace/hypertrace-ui/issues/1061

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -5,5 +5,5 @@ ignore:
   SNYK-JAVA-IONETTY-1042268:
     - '*':
         reason: None Given
-        expires: 2021-10-31T00:00:00.000Z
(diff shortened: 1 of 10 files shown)


aaron-steinfeld:
@hypertrace/query-layer-owners bump

kotharironak (approved):

PR #150, opened 2021-12-10
fix: use requested alias for result metadata

aaron-steinfeld:
## Description
In prior iterations EQS disregarded the alias being provided when naming its response columns, which is kind of the whole point of an alias :). This change fixes that.

### Testing
Added UT, Verified still working E2E.

### Checklist:


entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -522,4 +505,20 @@ public void total(
...
+
+  private ResultSetMetadata buildMetadataForSelections(List<Expression> selections) {
+    return ResultSetMetadata.newBuilder()
+        .addAllColumnMetadata(
+            () ->
+                selections.stream()
+                    .map(Expression::getColumnIdentifier)
+                    .map(
+                        identifier ->
+                            identifier.getAlias().isEmpty()
```
sarthak77, line 517: Can you add a test that would fail earlier but will pass now?

sarthak77:
Since this is a small PR, can we also fix `snyk` in this PR ?

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
aaron-steinfeld: yep just wanted to get it up, will add a test and fix snyk.

aaron-steinfeld: test added

sarthak77 (approved):

aaron-steinfeld:
@hypertrace/query-layer-owners please review.

kotharironak (approved):

PR #155, opened 2022-01-17
Implement converters to convert to the new Query DTO

suresh-prakash:
## Description
* Converters implementation of the proto to the new Query DTO class. This optimises the 2 phase conversion into a single phase
* Upgrade the document-store and the protobuf versions
* Introduces the Google Guice injection dependency

### Testing
Will add tests after one round of review

### Checklist:

### Documentation


entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/entity_query_request.proto
```diff
@@ -15,6 +15,8 @@ message EntityQueryRequest {
   repeated OrderByExpression orderBy = 7;
   int32 limit = 5;
   int32 offset = 6;
+  repeated GroupByExpression groupBy = 8;
+  // TODO: Do we need to support filtering on grouped values (HAVING clause)?
```
suresh-prakash, line 19: Do we have a use-case for `HAVING` clause, @skjindal93 ?

skjindal93: We always have a use-case. 

But, we don't support HAVING clause in query-service yet too. What does it take to add `HAVING` clause support in entity-service?

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/accessor/IOneOfAccessor.java
```diff
@@ -0,0 +1,11 @@
+package org.hypertrace.entity.query.service.converter.accessor;
+
+import java.util.Set;
+import org.hypertrace.entity.query.service.converter.ConversionException;
+
+public interface IOneOfAccessor<T, U extends Enum<U>> {
```
skjindal93, line 6: As a practice, we don't prefix interfaces with `I` 
```suggestion
public interface OneOfAccessor<T, U extends Enum<U>> {
```

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/accessor/OneOfAccessor.java
```diff
@@ -0,0 +1,46 @@
...
+  public <V> V access(final T proto, final U valueCaseEnum, final Set<U> allowedValueCases)
+      throws ConversionException {
+    if (!allowedValueCases.contains(valueCaseEnum)) {
+      throw new ConversionException(String.format("%s in %s is unsupported", valueCaseEnum, proto));
+    }
+
+    return access(proto, valueCaseEnum);
+  }
+
+  protected abstract void populate();
```
skjindal93, line 45: Can we rather just have a function, which returns a map? The map can then be used to populate the values in the class here

entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/entity_query_request.proto
suresh-prakash: Well, the attribute names used in the filter has to be consumed directly (instead of getting the sub-doc path).
So, it is quite some work to implement and test. My opinion is that if there is a strong requirement, we can implement now. Otherwise, we can defer.

Note: document-store already supports `HAVING` clause.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/filter/RelationalExpressionConverter.java
```diff
@@ -0,0 +1,95 @@
...
+    final Map<Operator, RelationalOperator> map = new EnumMap<>(Operator.class);
+
+    // TODO: Add support for NOT
+    //        map.put(NOT, RelationalOperator.NOT);
+    map.put(EQ, RelationalOperator.EQ);
+    map.put(NEQ, RelationalOperator.NEQ);
+    map.put(IN, RelationalOperator.IN);
+    map.put(NOT_IN, RelationalOperator.NOT_IN);
+    // TODO: Add support for range
+    //        map.put(RANGE, RelationalOperator.RANGE);
```
skjindal93, line 84:
```suggestion
    // map.put(RANGE, RelationalOperator.RANGE);
```

skjindal93 (commented):
Are the changes hooked in yet with the EntityQueryService APIs?

entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/entity_query_request.proto
skjindal93: Let's defer definitely. We might have to enhance `FilterConverter` later based on normal vs aggregation filtering

suresh-prakash: Sure.

suresh-prakash:
> Are the changes hooked in yet with the EntityQueryService APIs?

Not yet. Planning to have it in a separate PR in order to accommodate writing tests in this PR.

entity-service-impl/build.gradle.kts
```diff
@@ -31,6 +33,9 @@ dependencies {
   testImplementation("org.mockito:mockito-junit-jupiter:3.8.0")
   testRuntimeOnly("org.apache.logging.log4j:log4j-slf4j-impl:2.17.1")
 }
+repositories {
```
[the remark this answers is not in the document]
suresh-prakash: Accidentally, auto-added by IntelliJ, while I was upgrading the library versions. Removed.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/ConverterModule.java
```diff
@@ -0,0 +1,50 @@
...
+
+    bind(new TypeLiteral<Converter<LiteralConstant, ConstantExpression>>() {})
+        .to(ConstantExpressionConverter.class);
+    bind(new TypeLiteral<Converter<List<OrderByExpression>, Sort>>() {}).to(OrderByConverter.class);
+    bind(new TypeLiteral<Converter<EntityQueryRequest, Query>>() {}).to(QueryConverter.class);
+  }
+
+  @Provides
+  EntityAttributeMapping getEntityAttributeMapping() {
+    return attributeMapping;
```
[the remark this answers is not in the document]
suresh-prakash: We need to pass it to the constructor of ConverterModule.

skjindal93 (approved):

PR #156, opened 2022-01-20
feat(entity): created time field

skjindal93:


entity-service-api/src/main/proto/org/hypertrace/entity/data/service/v1/entity.proto
```diff
@@ -42,11 +42,14 @@ message Entity {
...
-  //attributes that uniquely identify the entity
+  // attributes that uniquely identify the entity
   map<string, AttributeValue> identifying_attributes = 6;
 
-  //attributes that describe the entity (e.g. service labels, service version, lang etc)
+  // attributes that describe the entity (e.g. service labels, service version, lang etc)
   map<string, AttributeValue> attributes = 7;
+  // no need to set it. document-store automatically adds a `createdTime` field
+  // can directly read the value from the returned entity
+  int64 created_time = 8;
```
aaron-steinfeld, line 52: please include unit
```suggestion
  int64 created_time_ms = 8;
```

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -230,12 +230,21 @@ Row convertToEntityQueryResult(
                           .setValueType(ValueType.STRING)
                           .setString(entity.getEntityName())
                           .build());
+                } else if (edsSubDocPath.equals(EntityServiceConstants.ENTITY_CREATED_TIME)) {
```
aaron-steinfeld, line 233: so if edsSubDocPath is equal to this, that means we put in an explicit ES mapping between some attribute ID and this path?

aaron-steinfeld: discussed offline, some mapping is required so reasonable to use the subdocpath map which is used for v1 entity id and names. V2 treats these as attributes as they're determined by the writer, but should already support using the mapping for this case.

An ideal but more difficult solution in terms of compatibility would be to store the creation timestamp attribute id on the entity types and then only write to there, but out of scope for now.

entity-service-api/src/main/proto/org/hypertrace/entity/data/service/v1/entity.proto
aaron-steinfeld: discussed offline - handcuffed here since this is an existing field in the doc already. Disregard.

aaron-steinfeld (approved):

PR #176, opened 2022-03-29
fix(entity-service): null filters as non existence of attribute

skjindal93:


suresh-prakash (dismissed):

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/filter/PrimitiveFilteringExpressionConverter.java
```diff
@@ -55,10 +57,44 @@ public FilterTypeExpression convert(
...
         constantExpressionConverter.convert(constant, requestContext);
 
     return RelationalExpression.of(identifierExpression, relationalOperator, constantExpression);
   }
+
+  private RelationalOperator normalizeOperator(Operator operator, Value value)
+      throws ConversionException {
+    Operator normalizedOperator = operator;
+
+    // A null value filter translates to non-existence of attribute
```
aaron-steinfeld, line 71: We never write nulls? I would think we'd want to translate `attr = null` to `(attr = null) OR (attr not exists)`, likewise on `!=`.

skjindal93: How do you even write nulls via proto APIs? Unless the proto API implementation adds `null` value directly to the entities

That being said, it still makes sense to translate `attr = null` to `(attr = null) OR (attr not exists)` as you mentioned

aaron-steinfeld: it could be written as string "null" for example - it depends on the calling code.

suresh-prakash (approved):

suresh-prakash (approved):

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/filter/NullFilteringExpressionConverter.java
```diff
@@ -0,0 +1,92 @@
...
+import org.hypertrace.entity.query.service.converter.identifier.IdentifierConverterFactory;
+import org.hypertrace.entity.query.service.v1.ColumnIdentifier;
+import org.hypertrace.entity.query.service.v1.LiteralConstant;
+import org.hypertrace.entity.query.service.v1.Operator;
+import org.hypertrace.entity.query.service.v1.Value;
+import org.hypertrace.entity.query.service.v1.ValueType;
+
+@Singleton
+@AllArgsConstructor(onConstructor_ = {@Inject})
+public class NullFilteringExpressionConverter extends FilteringExpressionConverterBase {
```
aaron-steinfeld, line 31: test?

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/filter/NullFilteringExpressionConverter.java
```diff
@@ -0,0 +1,92 @@
...
+import org.hypertrace.entity.query.service.converter.identifier.IdentifierConversionMetadata;
+import org.hypertrace.entity.query.service.converter.identifier.IdentifierConverter;
+import org.hypertrace.entity.query.service.converter.identifier.IdentifierConverterFactory;
+import org.hypertrace.entity.query.service.v1.ColumnIdentifier;
+import org.hypertrace.entity.query.service.v1.LiteralConstant;
+import org.hypertrace.entity.query.service.v1.Operator;
+import org.hypertrace.entity.query.service.v1.Value;
+import org.hypertrace.entity.query.service.v1.ValueType;
+
+@Singleton
```
aaron-steinfeld, line 29: Why singleton? Doesn't appear stateful

skjindal93: copy paste from other classes

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/ValueHelper.java
```diff
@@ -81,6 +83,10 @@ public boolean isMap(final ValueType valueType) {
     return MAP_TYPES.contains(valueType);
   }
 
+  public boolean isNull(final Value value) {
+    return NULL_VALUE.equalsIgnoreCase(value.getString());
```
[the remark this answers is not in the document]
skjindal93: What's the use case of the null case in QS? Who sets the null case value type?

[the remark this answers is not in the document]
aaron-steinfeld: Callers looking for things to not be null, for example :) Previously the only way to do that was the magic string "null", and that means we can't have a value actually named null. Also problematic for fields that aren't typed string. Put slightly differently, the purpose was to avoid putting implementation defaults/details on the caller.

suresh-prakash (dismissed):

entity-service-impl/src/test/java/org/hypertrace/entity/query/service/converter/filter/NullFilteringExpressionConverterTest.java
```diff
@@ -0,0 +1,129 @@
...
+
+  @Mock private EntityAttributeMapping entityAttributeMapping;
+  @Mock private IdentifierConverterFactory identifierConverterFactory;
+  @Mock private Converter<LiteralConstant, ConstantExpression> constantExpressionConverter;
+
+  private NullFilteringExpressionConverter nullFilteringExpressionConverter;
+
+  @BeforeEach
+  void setup() {
+    MockitoAnnotations.initMocks(this);
```
aaron-steinfeld, line 45: nit: use Mockito extension instead. Not only it do this automatically, but it sets other defaults that we want like strict mode too.

entity-service-impl/src/test/java/org/hypertrace/entity/query/service/converter/filter/NullFilteringExpressionConverterTest.java
```diff
@@ -0,0 +1,129 @@
...
+            columnIdentifier, Operator.EQ, constant, REQUEST_CONTEXT);
+    IdentifierExpression identifierExpression = IdentifierExpression.of("attributes.subDocPath1");
+
+    assertEquals(
+        LogicalExpression.builder()
+            .operator(LogicalOperator.OR)
+            .operands(
+                List.of(
+                    RelationalExpression.of(
+                        identifierExpression, RelationalOperator.NOT_EXISTS, constantExpression),
```
aaron-steinfeld, line 84: there shouldn't be a RHS of a not exists expression - is this being used or just a bad API?

entity-service-impl/src/test/java/org/hypertrace/entity/query/service/converter/filter/NullFilteringExpressionConverterTest.java
```diff
@@ -0,0 +1,129 @@
...
+            columnIdentifier, Operator.NEQ, constant, REQUEST_CONTEXT);
+    IdentifierExpression identifierExpression = IdentifierExpression.of("attributes.subDocPath1");
+
+    assertEquals(
+        LogicalExpression.builder()
+            .operator(LogicalOperator.OR)
+            .operands(
+                List.of(
+                    RelationalExpression.of(
+                        identifierExpression, RelationalOperator.EXISTS, constantExpression),
```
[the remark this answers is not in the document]
skjindal93: > If it exists and is set to null, this would incorrectly return true

I was assuming that we never set it to null. But, I also don't see a use case, we doing a `NEQ null` filter

entity-service-impl/src/test/java/org/hypertrace/entity/query/service/converter/filter/NullFilteringExpressionConverterTest.java
skjindal93: Just a bad API. It isn't being used on the document store implementation side

suresh-prakash (approved):

aaron-steinfeld (approved):

aaron-steinfeld:
Would you mind fixing deps in a followup PR?

entity-service-impl/src/test/java/org/hypertrace/entity/query/service/converter/filter/NullFilteringExpressionConverterTest.java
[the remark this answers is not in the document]
skjindal93: I just realized we actually have a use case for `NEQ null` filter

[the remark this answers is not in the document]
skjindal93: https://github.com/hypertrace/entity-service/pull/177

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/ValueHelper.java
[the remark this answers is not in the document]
skjindal93: I think I have hit this. Basically, the caller (gateway service) asks for say column1, column2, and there is a filter on `column2 != null`, the callee(entity service) would to still populate in some values for column2's `null` row

Currently, we put an empty string https://github.com/hypertrace/entity-service/blob/9e4cfdc4ff0b1e6bb5e68cf90f567e1a977ee193/entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/response/DocumentConverterImpl.java#L43 in entity-service

How should the `NEQ` null filtering happen in that case? The caller (gateway service) would have to take care of the `NEQ` null filtering in that case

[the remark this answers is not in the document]
aaron-steinfeld: I'm not quite sure I'm following - if there is a filter provided from GW on NEQ null, why are any results being returned that require populating a default value for null? There should never be a case where the caller has to do any filtering (so if there is, let's identify them)

PR #177, opened 2022-04-04
fix: handle neq null

skjindal93:


suresh-prakash (dismissed):

suresh-prakash (dismissed):

suresh-prakash (approved):

suresh-prakash (approved):

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/filter/NullFilteringExpressionConverter.java
```diff
@@ -74,6 +74,16 @@ public FilterTypeExpression convert(
             .operands(List.of(nonExistenceRelationalExpression, relationalExpression))
             .build();
 
+      case NEQ:
+        // 'field' NEQ 'null' -> 'field' NEQ 'null' && 'field' EXISTS
```
aaron-steinfeld, line 78: so if field exists, there's no change. If field is missing, what was NEQ null previously returning? I would have assumed false, which is the same as it would after this change.

So this logic doesn't seem wrong, just unnecessary unless mongo does some unexpected stuff.

skjindal93: https://www.mongodb.com/docs/manual/reference/operator/query/ne/

> [$ne](https://www.mongodb.com/docs/manual/reference/operator/query/ne/#mongodb-query-op.-ne) selects the documents where the value of the field is not equal to the specified value. This includes documents that do not contain the field.

We only want to apply the operator only if the field exists

aaron-steinfeld: Got it

aaron-steinfeld (approved):

PR #178, opened 2022-04-27
fix(entity-query-service): null filters as existence only filters

skjindal93:


skjindal93:
will fix snyk failures in a separate PR

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/filter/NullFilteringExpressionConverter.java
```diff
@@ -65,24 +62,14 @@ public FilterTypeExpression convert(
...
       case NEQ:
-        // 'field' NEQ 'null' -> 'field' NEQ 'null' && 'field' EXISTS
-        RelationalExpression existenceRelationalExpression =
-            RelationalExpression.of(
-                identifierExpression, convertOperator(Operator.EXISTS), constantExpression);
-        return LogicalExpression.builder()
-            .operator(LogicalOperator.AND)
-            .operands(List.of(existenceRelationalExpression, relationalExpression))
-            .build();
+        // 'field' NEQ 'null' -> 'field' EXISTS
```
avinashkolluru, line 70: so the assumption here is that if a field is set, it cant be null?

skjindal93: right

avinashkolluru (approved):

aaron-steinfeld (approved):

PR #182, opened 2022-06-28
chore | ENG-18747 | adding the delete via filter api to entity service

aman-bansal:


entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/entity_query_request.proto
```diff
@@ -69,3 +69,15 @@ message BulkEntityArrayAttributeUpdateRequest {
 
 message BulkEntityArrayAttributeUpdateResponse {
 }
+
+message DeleteEntitiesRequest {
```
aaron-steinfeld, line 73: Entity query service should be read-only (that's the "query" part 😉 ) . Please use or extend if necessary the EDS APIs.

aman-bansal: hmmm interesting, I was told the difference is based on treating it as generic objects or primary object. Also entity query service already has update, bulk update apis.

aaron-steinfeld: both services are generic. The update APIs in EQS are a mistake (for some reason we didn't want to call EDS from gateway service, so we duplicated the APIs there).

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
```diff
@@ -205,4 +207,9 @@ public void upsertEnrichedEntities(String tenantId, EnrichedEntities enrichedEnt
   public void upsertRelationships(String tenantId, EntityRelationships relationships) {
     client.upsertRelationships(tenantId, relationships);
   }
+
+  @Override
+  public DeleteEntitiesResponse deleteEntities(String tenantId, DeleteEntitiesRequest request) {
+    return client.deleteEntities(tenantId, request);
```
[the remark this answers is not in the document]
aman-bansal: how would we delete the cache based on filter?

entity-service-impl/src/main/java/org/hypertrace/entity/service/util/DocStoreConverter.java
```diff
@@ -53,6 +54,30 @@ public static <T extends GeneratedMessageV3> JSONDocument transform(T entity) th
     return new JSONDocument(json);
   }
 
+  public static org.hypertrace.core.documentstore.Filter transform(
+      @Nonnull String tenantId, @Nonnull DeleteEntitiesRequest request) {
+    List<Filter> filters = new ArrayList<>();
+    filters.add(getTenantIdEqFilter(tenantId));
+    if (StringUtils.isNotEmpty(request.getEntityType())) {
```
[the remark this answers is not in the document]
aman-bansal: i will make this necessary field.

entity-service-api/src/main/proto/org/hypertrace/entity/data/service/v1/entity_data_request.proto
```diff
@@ -100,3 +100,18 @@ message MergeAndUpsertEntityRequest {
 message MergeAndUpsertEntityResponse {
   Entity entity = 1;
 }
+
+message DeleteEntitiesRequest {
+  string entity_type = 1;
+  oneof delete_request {
+    StringList entity_ids = 2;
```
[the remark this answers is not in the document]
aman-bansal: Filter is an attribute filter. we don't push api id as attribute in api entity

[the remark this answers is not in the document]
aman-bansal: Otherwise i would have to add special handling for this logic in transform(AttributeFilter) function

entity-service-api/src/main/proto/org/hypertrace/entity/data/service/v1/entity_data_request.proto
```diff
@@ -100,3 +100,18 @@ message MergeAndUpsertEntityRequest {
...
+    AttributeFilter filter = 3;
+  }
+}
+
+message StringList {
+  repeated string ids = 1;
+}
+
+message DeleteEntitiesResponse {
```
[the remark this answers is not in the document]
aman-bansal: How would we manage the response size? Based on filters it could mean deleting lots of entities. To return the ids we would need to fetch all the entity ids based on the filter (some pagination logic) and then return (need to consider the response size). What do you suggest?

entity-service-api/src/main/proto/org/hypertrace/entity/data/service/v1/entity_data_request.proto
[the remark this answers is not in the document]
aaron-steinfeld: ugh, right. I hate entity service

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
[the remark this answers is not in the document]
aaron-steinfeld: https://github.com/hypertrace/entity-service/pull/182#discussion_r909569812 use the returned IDs

entity-service-api/src/main/proto/org/hypertrace/entity/data/service/v1/entity_data_request.proto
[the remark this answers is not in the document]
skjindal93: what if we add the delete API in entity query service, instead of entity data service. We do support ID as an attribute there. And, the delete API can then just work out of `entity_type` and attribute filter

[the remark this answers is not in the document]
aaron-steinfeld: https://github.com/hypertrace/entity-service/pull/182#discussion_r908418106

entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/entity_query_request.proto
skjindal93: > The update APIs in EQS are a mistake (for some reason we didn't want to call EDS from gateway service, so we duplicated the APIs there).

The update APIs in EQS also provides a way to directly update a specific attribute, and hence directly updating the specific field in mongo, instead of getting the entire object, and then updating the entire object, leading to memory pressure

EDS in my opinion always works with the entire entity object

skjindal93: I rather treat EQS as entity service, which works generically for attributes, and not only being used for querying, but can be used to update/delete attributes

Still, not convinced, why we shouldn't have update/delete APIs in EQS

skjindal93: > I could buy the argument that there shouldn't be a separate EDS and EQS just two different query APIs (one for entity objects, one for "rows" that supports aggregations)

Does it really matter then, where we keep the APIs? If the API design and implementation fits well with the EQS APIs, why not keep the APIs there? (for implementation simplicity)

For example: In case of delete API, the API could have been simple, by just working on the attributes in EQS, rather than having a one-of in EDS, because of the complex "id" transformation handling needed in EDS

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -543,9 +555,78 @@ public void total(
...
+      DeleteEntitiesRequest request, StreamObserver<DeleteEntitiesResponse> responseObserver) {
+    RequestContext requestContext = RequestContext.CURRENT.get();
+    if (validateDeleteEntitiesRequest(request, requestContext)) {
+      responseObserver.onError(new ServiceException("Invalid request provided"));
+      return;
+    }
+
+    List<String> entityIds = getEntityIdsToDelete(requestContext, request);
+    if (entityIds.size() == 0) {
+      LOG.info("{}. No entities found to delete", request);
```
aaron-steinfeld, line 569: if we're treating this as a success response, should make this a debug log. Debatable whether it should error out since this is likely an unintended request.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -543,9 +555,78 @@ public void total(
...
+      responseObserver.onNext(DeleteEntitiesResponse.newBuilder().build());
+      responseObserver.onCompleted();
+      return;
+    }
+
+    if (entityIds.size() > maxEntitiesToDelete) {
+      LOG.info(
```
aaron-steinfeld, line 576: warn log for error response.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -543,9 +555,78 @@ public void total(
...
+          "{}. Number of ids to delete exceeds the maximum limit: Entity Ids limit exceeded",
+          request);
+      responseObserver.onError(new RuntimeException("Entity Ids limit exceeded"));
+      return;
+    }
+
+    if (LOG.isDebugEnabled()) {
```
aaron-steinfeld, line 583: redundant if check, the logger knows if debug is enabled.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -543,9 +555,78 @@ public void total(
...
+      LOG.debug("Deleting entity of type: {} and ids: {}", request.getEntityType(), entityIds);
+    }
+
+    Optional<String> tenantId = requestContext.getTenantId();
+    this.entitiesCollection.delete(
+        DocStoreConverter.transform(tenantId.get(), request.getEntityType(), entityIds));
```
aaron-steinfeld, line 589: nit: prefer `.orElseThrow()` instead of `.get()`

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -543,9 +555,78 @@ public void total(
...
+      responseObserver.onError(new RuntimeException("Entity Ids limit exceeded"));
+      return;
+    }
+
+    if (LOG.isDebugEnabled()) {
+      LOG.debug("Deleting entity of type: {} and ids: {}", request.getEntityType(), entityIds);
+    }
+
+    Optional<String> tenantId = requestContext.getTenantId();
+    this.entitiesCollection.delete(
```
aaron-steinfeld, line 588: Always wrap the whole thing in a try/catch and report the caught error. If this errors out, currently it would result in a hanging request rather than reporting that back to the caller.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -543,9 +555,78 @@ public void total(
...
+    responseObserver.onCompleted();
+  }
+
+  private List<String> getEntityIdsToDelete(
+      RequestContext requestContext, DeleteEntitiesRequest request) {
+    List<Expression> selections =
+        List.of(
+            Expression.newBuilder()
+                .setColumnIdentifier(
+                    ColumnIdentifier.newBuilder().setColumnName(ENTITY_ID_ATTRIBUTE_ID).build())
```
aaron-steinfeld, line 600: bug - bad assumption that we only support api deletion.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -543,9 +555,78 @@ public void total(
...
+                .build());
+    EntityQueryRequest entityQueryRequest =
+        EntityQueryRequest.newBuilder()
+            .setEntityType(request.getEntityType())
+            .setFilter(request.getFilter())
+            .build();
+    Query query = entityQueryConverter.convertToEDSQuery(requestContext, entityQueryRequest);
```
aaron-steinfeld, line 607: I'm guessing this logic is duped from the query API. Once we map the delete to a query request, should be able to share the query and response processing logic.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -650,4 +731,19 @@ private Converter<List<Expression>, Selection> getSelectionConverter() {
                 Converter<
                     EntityQueryRequest, org.hypertrace.core.documentstore.query.Query>>() {}));
   }
+
+  private boolean validateDeleteEntitiesRequest(
+      DeleteEntitiesRequest request, RequestContext requestContext) {
+    Optional<String> tenantId = requestContext.getTenantId();
+    if (tenantId.isEmpty()) {
+      LOG.error("{}. Invalid deleteEntities request: Tenant id is not provided", request);
```
aaron-steinfeld, line 739: instead of logging an error and return a boolean, instead throw and let the caller decide what to do in case of an error (it should be returning that error to the caller and logging it at the top level).

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -650,4 +731,19 @@ private Converter<List<Expression>, Selection> getSelectionConverter() {
...
+      return true;
+    }
+
+    if (StringUtils.isEmpty(request.getEntityType())) {
```
aaron-steinfeld, line 743: nit: unnecessary util. getEntityType().isEmpty() (or isBlank())

aaron-steinfeld (changes requested):

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
aman-bansal: Could be that services might be sending just the filters without knowing if there are valid entities or not. Empty response is still valid response in those cases

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
aman-bansal: Interesting. I have seen this check everywhere across the code base. Maybe the difference in the library. I will check the implementation detail.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
aman-bansal: It's not precisely duped. It's taking some bits but not all. The problem is query Api requires selection as a list of Id attributes defined in the attributes map. Being generic we don't have that while deleting entities. So i am skipping the attribute mapping phase of result creation. Let me know if there is any other way?

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
aaron-steinfeld: That's what I was calling debatable. If the caller doesn't know it was valid or not, then an error told him it's invalid - the same as it would if it tried to delete by ID and that ID did not exist. It could be debated either way, so it's more important to be consistent with other APIs.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
aaron-steinfeld: Mostly tech debt, I clean it when I see it. It's from a misunderstanding how logging works.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
aaron-steinfeld: > The problem is query Api requires selection as a list of Id attributes defined in the attributes map

Not sure I'm following - you need the attributes to filter anyway, right? So I would assume you'd need to pull these from config. If you can get away without having to configure them, then great, but it just means a different unit would be shareable (the part up until result mapping).

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -559,66 +558,59 @@ public void total(
   public void deleteEntities(
       DeleteEntitiesRequest request, StreamObserver<DeleteEntitiesResponse> responseObserver) {
     RequestContext requestContext = RequestContext.CURRENT.get();
-    if (validateDeleteEntitiesRequest(request, requestContext)) {
-      responseObserver.onError(new ServiceException("Invalid request provided"));
-      return;
+    try {
+      validateDeleteEntitiesRequest(request, requestContext);
+    } catch (Exception ex) {
+      responseObserver.onError(ex);
```
aaron-steinfeld, line 564: return after reporting error. Currently will continue to execute.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -732,18 +724,18 @@ private Converter<List<Expression>, Selection> getSelectionConverter() {
...
   }
 
-  private boolean validateDeleteEntitiesRequest(
+  private void validateDeleteEntitiesRequest(
       DeleteEntitiesRequest request, RequestContext requestContext) {
     Optional<String> tenantId = requestContext.getTenantId();
     if (tenantId.isEmpty()) {
       LOG.error("{}. Invalid deleteEntities request: Tenant id is not provided", request);
-      return true;
+      throw new IllegalArgumentException(
```
aaron-steinfeld, line 732: Use GRPC exceptions - `Status.INVALID_ARGUMENT.withDescription("...").asRuntimeException()`

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
aman-bansal: EDS coupling is anyways there in execute api of EQS as well as total api. The difference delete api had is in the response translation. EQS other apis translate responses based on attributes provided, but we don't need that here. So that's why the difference. But I am sensing based on your comment that, this distinction was also mixed sometime before. let me try the without this mapping

aman-bansal: > Couldn't you just send a query to EDS by translating the filters and then getting the response entities and taking the IDs out of that? The usage of the constant is only necessary because you're directly calling doc store it looks like?

EQS doesn't call EDS directly. I don't think we should add that. We can only use EQS `execute` api only but my concern is around how to add the correct selection filter for entityId. Because those `execute` assumed that they will from the attribute map only. Can I assume the entity id attribute will always be `{ENTITY_TYPE}.id`. I don't see anything that works on this.

aaron-steinfeld: > EQS doesn't call EDS directly. I don't think we should add that.

It can and really should - we're saying we'd prefer to couple their implementations (e.g. sharing constants and duplicating logic and have two different services reading and writing the same data) rather than than having one data owner and another that presents a different API to the same data? That's a majorly problematic design. I'm not saying we need to fix that everywhere, we've already dug ourselves in deep, but if we're avoiding that as an easy solution to the problem we need not be.

> Can I assume the entity id attribute will always be {ENTITY_TYPE}.id.

No - if we don't want to rely on EDS (see above, we should be!), we'll need to maintain a map similar to gateway service https://github.com/hypertrace/gateway-service/blob/main/gateway-service/src/main/resources/configs/common/application.conf#L53

skjindal93: Why aren't we using the other `queryConverter.convert(request, requestContext);` we have in the query implementation?

https://github.com/hypertrace/entity-service/blob/c1ab8baf0077c92d8c5aa05e92507f55b3620f31/entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java#L157

We are planning to remove the entire `convertToEDSQuery` method, since it was deprecated (since aggregation support didn't exist in Postgres before)

suresh-prakash: Yes. Ronak is working on implementing the new Query APIs for Postgres and is expected to complete by this week. Most of the functionality is already there. And, the functionality needed for this purpose is already supported.
You may have to cut a latest release of document-store and use it here.

If everything goes well, we will have the `queryAggregationEnabled` flag removed soon, defaulting it to `true`. So, instead of surrounding this block I'd suggest that we use the new (document-store) query API (that directly converts the EQS query to document store query without the intermediate EDS query conversion). Better to get an approval from @kotharironak 

You could do some refactoring to extract out the common code block between `execute` and this method.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -543,9 +555,72 @@ public void total(
...
+    }
+  }
+
+  private List<String> getEntityIdsToDelete(
+      RequestContext requestContext, DeleteEntitiesRequest request) {
+    List<String> docStoreSelections = List.of(EntityServiceConstants.ENTITY_ID);
+    EntityQueryRequest entityQueryRequest =
+        EntityQueryRequest.newBuilder()
+            .setEntityType(request.getEntityType())
+            .setFilter(request.getFilter())
```
suresh-prakash, line 604: For optimisation purposes, we could add a limit `maxEntitiesToDelete + 1` since we know we are never going to consume more than `maxEntitiesToDelete`.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -543,9 +555,72 @@ public void total(
...
+            .build();
+    Query query = entityQueryConverter.convertToEDSQuery(requestContext, entityQueryRequest);
+    CloseableIterator<Document> documentIterator =
```
suresh-prakash, line 607: Close documentIterator in `finally` block to release resources.

suresh-prakash:
Since a new API endpoint is added, can we have add integration tests?

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
avinashkolluru: https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

suresh-prakash (approved):
Looks good to me.

PR #183, opened 2022-07-02
refactor: use service factory abstraction, health check

aaron-steinfeld:
## Description
Wrapped entity service in factory framework. Depends on https://github.com/hypertrace/service-framework/pull/43


skjindal93:
Curious, to know what's the service factory abstraction?

aaron-steinfeld:
> Curious, to know what's the service factory abstraction?

https://github.com/hypertrace/service-framework/tree/main/platform-grpc-service-framework/src/main/java/org/hypertrace/core/serviceframework/grpc

Basically just some common interfaces and base classes rather than copying the same container code to every repo. By implementing the factory framework, we can compose any number of grpc services in platform containers easily. This repo adds that factory, and changes this service to use the standalone version of the container with that factory.

surajpuvvada (approved):

PR #187, opened 2022-08-04
chore | skip change notifications in case of certain attributes

aman-bansal:


entity-service/src/main/resources/configs/common/application.conf
```diff
@@ -21,6 +21,28 @@ attribute.service.config = {
   port = ${?ATTRIBUTE_SERVICE_PORT_CONFIG}
 }
 
+entity.service.change.skip.attributes = [
+  {
+    scope = API
+    attributes = ["last_called_millis", "change_label_timestamp", "api_discovery_timestamp"]
+  },
+  {
+    scope = ACTOR
```
avinashkolluru, line 30: Get rid of this common config. It should get applied from helm/values as needed by the use case

entity-service-change-event-generator/src/main/java/org/hypertrace/entity/service/change/event/impl/EntityChangeEventGeneratorImpl.java
```diff
@@ -37,10 +47,19 @@ public class EntityChangeEventGeneratorImpl implements EntityChangeEventGenerato
...
   EntityChangeEventGeneratorImpl(Config appConfig, Clock clock) {
     Config config = appConfig.getConfig(EVENT_STORE);
+    this.entityToAttributeChangeSkipMap =
+        config.hasPath(SKIP_ATTRIBUTES_CONFIG_PATH)
+            ? config.getConfigList(SKIP_ATTRIBUTES_CONFIG_PATH).stream()
+                .collect(
+                    Collectors.toUnmodifiableMap(
+                        conf -> conf.getString(SCOPE_PATH),
+                        conf -> conf.getStringList(ATTRIBUTES_PATH)))
+            : new HashMap<>();
```
aaron-steinfeld, line 62: we don't want the map to be mutable in this path and immutable in the other - should always be immutable.

entity-service-change-event-generator/src/main/java/org/hypertrace/entity/service/change/event/impl/EntityChangeEventGeneratorImpl.java
```diff
@@ -147,6 +172,67 @@ private void sendUpdateNotification(
     }
   }
 
+  private boolean shouldSendNotification(Entity prevEntity, Entity currEntity) {
+    boolean isAttributesChanged = false, isCoreEntityChanged = false;
+    for (FieldDescriptor fieldDescriptor : Entity.getDescriptor().getFields()) {
```
aaron-steinfeld, line 177: Sorry, why are we messing with the proto here? The fields inside the attributes map should be all we care about.

entity-service-change-event-generator/src/main/java/org/hypertrace/entity/service/change/event/impl/EntityChangeEventGeneratorImpl.java
```diff
@@ -147,6 +172,67 @@ private void sendUpdateNotification(
...
+                        .get(entityType)
+                        .contains(attributeKey.getKey())))
+            .collect(Collectors.toList());
+
+    return addedAttributes.size() > 0
+        || deletedAttributes.size() > 0
+        || updatedAttributes.size() > 0;
+  }
+
+  private boolean isCoreEntityChanged(Entity prevEntity, Entity currEntity) {
```
aaron-steinfeld, line 228: What's a core entity? And do we even support changing an entity name (even if we do, the name should also be an attribute)

entity-service-change-event-generator/src/main/java/org/hypertrace/entity/service/change/event/impl/EntityChangeEventGeneratorImpl.java
```diff
@@ -147,6 +172,67 @@ private void sendUpdateNotification(
...
+    return !prevEntity.getEntityName().equals(currEntity.getEntityName());
+  }
+
+  private boolean shouldSkipAttribute(String entityType, String attributeKey) {
+    return this.entityToAttributeChangeSkipMap.get(entityType).contains(attributeKey);
```
aaron-steinfeld, line 233: this will NPE if it encounters an entity type that didn't happen to be in the config, like the default.

entity-service-change-event-generator/src/main/java/org/hypertrace/entity/service/change/event/impl/EntityChangeEventGeneratorImpl.java
```diff
@@ -147,6 +172,67 @@ private void sendUpdateNotification(
...
+          }
+          break;
+      }
+    }
+
+    return isCoreEntityChanged || isAttributesChanged;
+  }
+
+  private boolean checkIfAttributesChanged(Entity prevEntity, Entity currEntity) {
+    MapDifference<String, AttributeValue> attributesDiff =
```
aaron-steinfeld, line 196: This can dramatically simplify.
1. Remove the skipped attributes out of both maps before doing the diff, rather than doing it after each direction of diff.
2. We don't actually care what the diff is, just that there is one, so rather than taking added, deleted and changed, you can just say - `prevMap.size == currMap.size == entriesInCommon.size`

entity-service-change-event-generator/src/main/java/org/hypertrace/entity/service/change/event/impl/EntityChangeEventGeneratorImpl.java
aman-bansal: My assumption was first class params can also change. we should skip those cases too. Can I assume at least for entity cases that it will always be attributes that might change? Because of this only i have to play with proto fields descriptor that you mentioned above

aaron-steinfeld: I would just assume first class params can not change. That would open up a can of worms in terms of how to specify it etc. anyway, and since we shouldn't be using first class params in the first place (at least, anything written in the last 2 or so years), we don't need to implement functionality for it.

entity-service/src/main/resources/configs/common/application.conf
```diff
@@ -21,6 +21,12 @@ attribute.service.config = {
   port = ${?ATTRIBUTE_SERVICE_PORT_CONFIG}
 }
 
+entity.service.change.skip.attributes = [
+  {
+    scope = API
+    attributes = ["last_called_millis"]
```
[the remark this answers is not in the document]
aman-bansal: I wanted to do that initially, but if we map it to attribute ID then we have to add the dependency to the CachingAttributeClient to EntityChangeEventGenerator. Because then we need to resolve these attributes and their paths. Is there any other way to achieve this without CachingAttributeClient? Adding the dependency of AttributeClient seems overkill.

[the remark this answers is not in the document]
aaron-steinfeld: Not necessarily needed. The attribute client helps go from an attribute ID to its metadata - here, you need to go from an entity subpath to an attribute ID. EQS already has a map of data to do this in the other direction, and that's what you'd need access to.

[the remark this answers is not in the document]
aman-bansal: I dont think i understand this correctly. Lets say we define the skip map something like this
```
entity.service.change.skip.attributes = [
        {
               scope = API,
               attributeId = API.lastCalledTimestamp
               // or
              subDocPath = attributes.last_called_millis (though it's same as above just because of subdoc path we are adding attributes.)
        
        }
]
```

Now in case one, where we are defining attributeId, we have to integrate with attributesClient because we can't be sure that the given attribute will be part of the event service config. It can come from attribute service also. And EventGenerator needs the exact doc store path.
Now the second case with subDocPath is similar to what I have already defined. It's better to add the attribute name directly to skip some translation logic (to remove attributes from subdoc path).

Is my understanding correct? or am I going in a completely wrong direction here?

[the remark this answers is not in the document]
aaron-steinfeld: I'm suggesting here that we define it like:
```
// Just attribute ids
entity.service.change.skip.attributes = ["API.lastCalledTimestamp", "API.xyz"]
```

Now, when the change event generator is handed an entity, it can basically take this https://github.com/hypertrace/entity-service/blob/6f000f36d732bcb49ce60b00adc3afdbe71ee8a3/entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityAttributeMapping.java#L59-L63

And just added the inverse function - `getAttributeIdFromDocStorePath`. This should allow each value in the entity to be mapped to an attribute ID, which can then be checked against the list. No attribute lookup needed. There may be a few fields without attributes, but that's purely tech debt (because we stopped defining separate entity attributes years ago), so we should just default to not support skipping them and if we run into the scenario where we need to, it's forcing function to clean up tech debt.

Does that make sense? That's basically what I was thinking at a high level.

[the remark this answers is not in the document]
aman-bansal: `calculateDocStorePathFromAttributeId` actually makes a call to the attribute service Client. So if I understand this correctly, what you are suggesting is for each attribute in the entity, we will call `getAttributeIdFromDocStorePath` which will return list of AttributeIds and then filter out based on configuration present. 

But the problem still remains the same, even though there will be a few attributes that are not defined, it still needs to make the call to the attribute client to verify all the possible places where we can define the attributes.

If we really want to go with AttributeId, i think we should go with something like this:
```
// Just attribute ids
entity.service.change.skip.attributes = ["API.lastCalledTimestamp", "API.xyz"]

private EntityAttributeMapping attributeMapping
public EntityChangeEventGenerator(EntityAttributeMapping attributeMapping) {
attributeMapping = attributeMapping;
}

public void generateEvent(Entity entity) {
 attributePaths = skipAttributes.forEach(attribute -> attributeMapping.getDocStorePathByAttributeId(attribute)).collect(Collectors.toList());
// . . . 
Filter out the attributes from entity and then check the difference.
// . . . .
}
```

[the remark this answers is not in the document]
aaron-steinfeld: Discussed a bit offline, just 2 bits to add here:
- Making a call to attribute client shouldn't really be an issue if perf is your concern here. It's a caching client and already called anyway (assuming we have called EQS), so there's no remote calls being made here.
- the above is fine - whether we start with the subdoc path and translate it to an id, or give all skipped ids and translate to subdoc paths isn't a huge deal given the relevant quantities. Only thing missing there is that you've lost your concept of scope when you got those paths back (e.g. I mapped API.xyz to the subdoc path xyz, but for all I know the entity is of type service which _also_ has a subdoc path xyz that hasn't been mapped to skipping). So there's going to be a bit more to do, but not much.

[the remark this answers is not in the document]
aman-bansal: I am going ahead with the above approach. In this way, I don't have to translate all the available attributes in an entity.

entity-service-change-event-generator/src/main/java/org/hypertrace/entity/service/change/event/impl/EntityChangeEventGeneratorImpl.java
```diff
@@ -147,6 +166,33 @@ private void sendUpdateNotification(
...
   }
 
+  private boolean shouldSendNotification(
+      RequestContext requestContext, Entity prevEntity, Entity currEntity) {
+    String entityType = prevEntity.getEntityType();
+    Entity.Builder prevEntityBuilder = prevEntity.toBuilder();
+    Entity.Builder currEntityBuilder = currEntity.toBuilder();
+    this.changeNotificationSkipAttributeList.forEach(
+        attributeId -> {
+          if (!attributeId.startsWith(entityType)) {
```
aman-bansal, line 176: This is one way to define whether given attribute matches the entity type or not. The other way could be to define the config as an map of entity types vs attributeID. something similar we had before like this
```
{
    scope = API
    attributeIds = ["last_called_millis"]
  }
```
@aaron-steinfeld thoughts?

aaron-steinfeld: > {
    scope = API
    attributeIds = ["last_called_millis"]
  }

could be just the example (that's a subdoc path), but that's the whole thing we're trying to get away from. Just to clarify terminology:
```
scope: API
attribute key: lastCalledTimestamp
attribute ID: API.lastCalledTimestamp
subDocPath: last_called_millis (or maybe it was attributes.last_called_millis, can't remember)
```

Now when we get an entity object, we've got a scope and a subdocpath, and we want to get that to the common platform attribute model - either a scope + attr key, or attribute ID.

We know how to get the attribute ID for explicitly mapped ones from the other config used in EntityAttributeMapping. We also know how to reverse the logic for the implicitly mapped ones to get to an attribute key (and already have scope). So in one path we end up with an attribute ID, in the other path an attribute key. Either one is sufficient for config, but ideally we want them to be consistent of course.

So we could model as:
1. A list of IDs, use the attribute client to translate scope + key (in implicit case) into IDs
2. A map of <scope,key>, use the attribute client to translate IDs (in explicit map case) into scope, key
3. double up the config to hold the scope, key, and attribute ID - seems redundant.
4. Skip the attribute client and use the fact that `id = ${scope}.${key}` to translate for one or the other. I really suggest we don't do that since we've already got the client and this would be breaking encapsulation for no good reason.

entity-service-common/build.gradle.kts
```diff
@@ -0,0 +1,20 @@
+plugins {
```
aaron-steinfeld, line 1: Please no generic common libs. Make it specific (e.g. attribute translator)

entity-service-change-event-generator/src/main/java/org/hypertrace/entity/service/change/event/impl/EntityChangeEventGeneratorImpl.java
aman-bansal: oh shit! I copied the config and forgot to change the id. What I meant was something like this
```
{
    scope = API
    attributeIds = ["API.lastCalledTimestamp"]
  }
```

aaron-steinfeld: see above, doesn't give us a way, other than 4, to go from an implicit subdoc path to attribute ID.

entity-service-attribute-translator/src/main/java/org/hypertrace/entity/attribute/translator/AttributeMetadata.java
```diff
@@ -0,0 +1,16 @@
+package org.hypertrace.entity.attribute.translator;
+
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+@Data
```
aaron-steinfeld, line 6: Probably just @Value, since we don't want it mutable. Also both Value and Data include EqualsAndHashCode, can be removed.

entity-service-attribute-translator/src/main/java/org/hypertrace/entity/attribute/translator/AttributeMetadata.java
```diff
@@ -0,0 +1,16 @@
...
+@EqualsAndHashCode
+public class AttributeMetadata {
```
aaron-steinfeld, line 8: Naming: AttributeMetadata is the name of what''s being represented, this is just an identifier. Maybe `AttributeIdentifier` or `AttributeMetadataIdentifier`?

entity-service-change-event-generator/src/main/java/org/hypertrace/entity/service/change/event/impl/EntityChangeEventGeneratorImpl.java
```diff
@@ -147,6 +168,33 @@ private void sendUpdateNotification(
...
   }
 
+  private boolean shouldSendNotification(
+      RequestContext requestContext, Entity prevEntity, Entity currEntity) {
+    String entityType = prevEntity.getEntityType();
+    Entity.Builder prevEntityBuilder = prevEntity.toBuilder();
+    Entity.Builder currEntityBuilder = currEntity.toBuilder();
+    this.changeNotificationSkipAttributeList.forEach(
+        attributeId -> {
+          Optional<AttributeMetadata> attributeScopeKey =
```
aaron-steinfeld, line 178: naming

entity-service-change-event-generator/src/main/java/org/hypertrace/entity/service/change/event/impl/EntityChangeEventGeneratorImpl.java
```diff
@@ -147,6 +168,33 @@ private void sendUpdateNotification(
...
+              this.entityAttributeMapping.getAttributeMetadataByAttributeId(
+                  requestContext, attributeId);
+          if (attributeScopeKey.isPresent()
```
aaron-steinfeld, line 181: nit: prefer ifPresent rather than explicit isPresent + get's (produces the same result, but get() requires logic, while ifPresent will give you compile-time validation).

entity-service-change-event-generator/src/main/java/org/hypertrace/entity/service/change/event/impl/EntityChangeEventGeneratorImpl.java
```diff
@@ -147,6 +168,33 @@ private void sendUpdateNotification(
...
+            currEntityBuilder.removeAttributes(attributeName);
+          }
+        });
+
+    MapDifference<String, AttributeValue> attributesDiff =
+        Maps.difference(
+            prevEntityBuilder.build().getAttributesMap(),
+            currEntityBuilder.build().getAttributesMap());
+
+    return !attributesDiff.areEqual();
```
aaron-steinfeld, line 195: nit unnecessary variable, just return directly.

entity-service-attribute-translator/src/main/java/org/hypertrace/entity/attribute/translator/AttributeMetadataIdentifier.java
```diff
@@ -0,0 +1,14 @@
+package org.hypertrace.entity.attribute.translator;
+
+import lombok.Value;
+
+@Value
+public class AttributeMetadataIdentifier {
+  String scope;
+  String docStorePath;
+
+  public AttributeMetadataIdentifier(String scope, String docStorePath) {
```
aaron-steinfeld, line 10: this exact constructor is generated by value

>`@Value` is shorthand for: `final @ToString` `@EqualsAndHashCode` `@AllArgsConstructor` `@FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE)` `@Getter`

aaron-steinfeld (approved):
A few minor variable naming issues after the renamed class.

aaron-steinfeld (approved):

PR #190, opened 2022-08-10
chore | generating change events from EQS

aman-bansal:


skjindal93 (approved):

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -353,7 +381,14 @@ private void doUpdate(RequestContext requestContext, EntityUpdateRequest request
         }
       }
       try {
+        List<Entity> existingEntities =
```
avinashkolluru, line 384: I think this part can be optimized. We can check for the attributes being updated and if they match the attributes configured in https://github.com/hypertrace/entity-service/pull/187/files we can bypass the entire change notification code path.

aman-bansal: We already have this change evaluation logic in entityChangeEventGenerator (Though it won't have any effect on scale here). So it will be bit of duplication. Adding here not sure how much it will help. The problem I feel is the scale of each EQS update. We are probably trying to update lots of entities and that is too frequently. I believe the easiest solution could be to enable it for some specific entity types only. We anyways don't need this for all entities i guess.

aaron-steinfeld: You can share the logic. If the issue is that certain things change very frequently and they're not worth sending notifications for (e.g. a timestamp), then this current logic makes things significantly worse. Rather than a single call to update, we are now forced to query twice regardless of what was updated and whether we'll send an event. We can look at the change request to determine whether we should change, and only fetch the entities if we know we'll need them.

PR #194, opened 2022-08-23
chore | generating change events from EQS

aman-bansal:


entity-service-attribute-translator/src/main/java/org/hypertrace/entity/attribute/translator/EntityAttributeChangeEvaluator.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.stream.Collectors;
+import org.hypertrace.core.grpcutils.context.RequestContext;
+import org.hypertrace.entity.data.service.v1.Entity;
+import org.hypertrace.entity.query.service.v1.ColumnIdentifier;
+import org.hypertrace.entity.query.service.v1.UpdateOperation;
+
+public class EntityAttributeChangeEvaluator {
+
+  private static final String SKIP_ATTRIBUTES_CONFIG_PATH = "entity.service.change.skip.attributes";
+  public static final String ALL_ENTITIES = "*";
```
skjindal93, line 18:
```suggestion
  public static final String ALL_ENTITY_TYPES = "*";
```

entity-service-attribute-translator/src/main/java/org/hypertrace/entity/attribute/translator/EntityAttributeChangeEvaluator.java
```diff
@@ -0,0 +1,120 @@
...
+  private final List<String> allowedEntityTypes;
+  private final List<String> changeNotificationSkipAttributeList;
+  private final EntityAttributeMapping entityAttributeMapping;
+
+  public EntityAttributeChangeEvaluator(
+      Config appConfig, EntityAttributeMapping entityAttributeMapping) {
+    this(appConfig, List.of(ALL_ENTITIES), entityAttributeMapping);
```
skjindal93, line 25: Can we change list into set?

entity-service-attribute-translator/src/main/java/org/hypertrace/entity/attribute/translator/EntityAttributeChangeEvaluator.java
```diff
@@ -0,0 +1,120 @@
...
+      EntityAttributeMapping entityAttributeMapping) {
+    this.changeNotificationSkipAttributeList = appConfig.getStringList(SKIP_ATTRIBUTES_CONFIG_PATH);
+    this.allowedEntityTypes = allowedEntityTypes;
+    this.entityAttributeMapping = entityAttributeMapping;
+  }
+
+  public boolean shouldSendNotification(
+      RequestContext requestContext, Entity prevEntity, Entity currEntity) {
+    String entityType = prevEntity.getEntityType();
+    if (!isEntityTypesAllowed(entityType)) {
```
skjindal93, line 40: nit:
```suggestion
    if (!isEntityTypeAllowed(entityType)) {
```

entity-service-attribute-translator/src/main/java/org/hypertrace/entity/attribute/translator/EntityAttributeChangeEvaluator.java
```diff
@@ -0,0 +1,120 @@
...
+
+  private String removePrefix(String str, final String prefix) {
+    if (str != null && prefix != null && str.startsWith(prefix)) {
+      return str.substring(prefix.length());
+    }
+    return str;
+  }
+
+  private boolean isEntityTypesAllowed(String entityType) {
+    return allowedEntityTypes.contains("*") || allowedEntityTypes.contains(entityType);
```
skjindal93, line 118:
```suggestion
    return allowedEntityTypes.contains(ALL_ENTITIES) || allowedEntityTypes.contains(entityType);
```

entity-service/src/main/resources/configs/common/application.conf
```diff
@@ -22,6 +22,7 @@ attribute.service.config = {
 }
 
 entity.service.change.skip.attributes = []
+eqs.change.notification.enabled.entity.types = ["*"]
```
skjindal93, line 25: We should have control for both EQS and EDS. May be even a single config control for both. Otherwise, it becomes inconsistent

We should also club entity change related configs into a single section. May be something like

```
entity.service.change {
  allowedEntityTypes = [`*`]
  skipAttributes = []
}
```

aman-bansal: I knew it this was coming 😂 . So the thing is I am not sure how much impact all this change will have on the scale. Feels like we need at least in the start more control over the behavior of EQS, otherwise, we might run into the same issue again. I checked the traffic over dogfooding and lots seem like time updates only. Let me change it. will monitor it more carefully when deployed

skjindal93 (approved):

PR #198, opened 2022-09-14
fix(filter): support filtering empty arrays

skjindal93:


entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/filter/NullFilteringExpressionConverter.java
```diff
@@ -31,8 +33,8 @@ public FilterTypeExpression convert(
       final LiteralConstant constant,
       final RequestContext requestContext)
       throws ConversionException {
-    final String id = columnIdentifier.getColumnName();
-    final String subDocPath = getSubDocPathById(entityAttributeMapping, id, requestContext);
+    final String columnId = columnIdentifier.getColumnName();
+    final String subDocPath = getSubDocPath(columnId, requestContext);
 
     final IdentifierExpression identifierExpression = IdentifierExpression.of(subDocPath);
```
suresh-prakash, line 39: Can be simplified after injecting org.hypertrace.entity.query.service.converter.identifier.IdentifierExpressionConverter. 
```suggestion
    final IdentifierExpression identifierExpression = identifierExpressionConverter.convert(columnIdentifier);
```

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/filter/NullFilteringExpressionConverter.java
```diff
@@ -45,16 +47,37 @@ public FilterTypeExpression convert(
...
+  /**
+   * An array is stored as 'attributes.array_field.valueList.values.%d.value.&lt;type&gt;'.
+   *
+   * <p>Example: products = ['shoes', 'socks'] is stored as
+   * 'attributes.names.valueList.values.0.value.string: shoes', and
+   * 'attributes.names.valueList.values.1.value.string: socks'.
+   *
+   * <p>An empty array is stored as 'attributes.array_field.valueList: {}'
+   */
+  private String getSubDocPath(String columnId, RequestContext requestContext)
```
suresh-prakash, line 74: Once the above comment is addressed, we could get rid of these 2 methods as the same business logic is being done in org.hypertrace.entity.query.service.converter.identifier.IdentifierExpressionConverter

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/filter/NullFilteringExpressionConverter.java
skjindal93: Yes. That's a nice suggestion!

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/filter/NullFilteringExpressionConverter.java
```diff
@@ -10,18 +8,16 @@
...
+import org.hypertrace.entity.query.service.converter.identifier.IdentifierExpressionConverter;
 import org.hypertrace.entity.query.service.v1.ColumnIdentifier;
 import org.hypertrace.entity.query.service.v1.LiteralConstant;
 import org.hypertrace.entity.query.service.v1.Operator;
 
 @AllArgsConstructor(onConstructor_ = {@Inject})
 public class NullFilteringExpressionConverter extends FilteringExpressionConverterBase {
-  private final EntityAttributeMapping entityAttributeMapping;
-  private final IdentifierConverterFactory identifierConverterFactory;
+  private final IdentifierExpressionConverter identifierExpressionConverter;
```
suresh-prakash, line 20: Is this binding working? I see that IdentifierExpressionConverter is bound to `Converter<ColumnIdentifier, IdentifierExpression>` only @ IdentifierModule

suresh-prakash (approved):

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/filter/NullFilteringExpressionConverter.java
skjindal93: fair point. let me test this end to end

skjindal93: tested this e2e. seems to be working fine

PR #210, opened 2022-11-15
Enhance bulk update to support updating by filters

suresh-prakash:
## Description
Previously, the bulk update entities API only supported updating by id. So, if a client wants to update all the entities matching certain criteria, it had to be done in 2 passes:
1. Get the ids of the entities matching the criteria (using the "execute" API)
2. Bulk update the entities using those ids (using the "bulkUpdate" API)

This approach has the following drawbacks for bulk updating entities based on some criteria/filter.
1. Verbosity: The client has to write more code
2. Performance: Two GRPC calls (and hence 2 database calls) need to be made
3. Concurrency: Any new entity created after the first call was made would not be updated by the second call

This PR enhances the "bulkUpdate" API to perform update by any attribute filter (not just by id) so that these drawbacks can be addressed.

### Testing
None of the existing tests are breaking

### Checklist:


entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/entity_query_request.proto
```diff
@@ -35,6 +35,13 @@ message BulkEntityUpdateRequest {
   }
 }
 
+message BulkUpdateEntitiesRequest {
+  // The query defining the results returned after update.
+  // Entity type and filter for update will be fetched from the query.
+  EntityQueryRequest query = 1;
+  repeated UpdateOperation operations = 2;
```
skjindal93, line 42: Do we lose a way now to specify different types of updates for different entities in a single request?

suresh-prakash: Yes. That can lead to timeout since that would involve making multiple calls to the database.

Moreover, it made sense to return the set of documents (by unioning the ids) in the older way.
But, with different filter criteria, what we return could become confusing if allow different updates based on different criteria.

suresh-prakash: Makes sense. Done.

aaron-steinfeld: can't add a oneof without breaking compatibility. Maybe we can come up with some clearer naming though (I mean in terms of conveying behavior - not sure if now is the time to start following our general proto best practices in this repo)? The API itself seems fine.

aaron-steinfeld: oh sorry - I meant on the method name - that it's unclear the difference between bulkUpdate and bulkUpdateEntities. On the best practices, I meant using appropriate casing, removing the streaming result (like line 23 _almost_ did - missed the casing on the method though).

In terms of this field... Putting a query in the update is a little bit weird. Do we intend on limit and offset being enforced? What if they put in a group by?

I think I would probably just give two separate fields - a filter and a selection for the results.

suresh-prakash: > that it's unclear the difference between bulkUpdate and bulkUpdateEntities

Updated it to `BulkUpdateAllMatchingFilter`.

> On the best practices, I meant using appropriate casing, removing the streaming result (like line 23 almost did - missed the casing on the method though).

Renamed the method name to be in title case and removed response streaming.

> Do we intend on limit and offset being enforced?

No. All will be updated at once.

> What if they put in a group by?

Was planning to add a validation in the implementation

> I think I would probably just give two separate fields - a filter and a selection for the results.

Yes. Just taking in the filter now and not returning any results as we can see that's what is being done by the current bulkUpdate API as well.

suresh-prakash (commented):
Commented the implementation for now. Will fix once the API is frozen.

entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/entity_query_request.proto
```diff
@@ -52,6 +62,19 @@ message SetAttribute {
...
+message AttributeUpdateOperation {
+  ColumnIdentifier attribute = 1;
+  AttributeUpdateOperator operator = 2;
+  LiteralConstant value = 3;
+
+  enum AttributeUpdateOperator {
+    ATTRIBUTE_UPDATE_OPERATION_UNSPECIFIED = 0;
+    ATTRIBUTE_UPDATE_OPERATOR_SET = 1;
+    ATTRIBUTE_UPDATE_OPERATOR_ADD_TO_LIST_IF_ABSENT = 2;
+    ATTRIBUTE_UPDATE_OPERATOR_REMOVE_ALL_FROM_LIST = 3;
```
skjindal93, line 74: What would be the `LiteralConstant value` in this case? Is it something like UNSET?

Should we instead do the following?

`SET` -> sets an attribute value. Can it also set an entire array attribute?
`UNSET` -> unsets an attribute value. Basically, removes the attribute. This wouldn't need literal constant being set.
`ADD_TO_LIST` -> adds multiple items to list
`REMOVE_FROM_LIST` -> removes multiple items from list

suresh-prakash: Definition of operators.
`SET` -> sets an attribute value. The value can either be a single value or a complete array.
`UNSET` -> not added here, but document store has support (in this PR: https://github.com/hypertrace/document-store/pull/145/files) and so we can add it if required. Note: For this, the literal value is optional and will be ignored.
`ADD_TO_LIST_IF_ABSENT` -> can either add single or multiple items to a list (add only those which are not already present to be consistent with the existing "bulkUpdateEntityArrayAttribute" API)
`ADD_TO_LIST` -> can directly add either a single or multiple items to a list -- We can add this support if required since document-store supports it (in this PR: https://github.com/hypertrace/document-store/pull/145/files)
`REMOVE_FROM_LIST` -> can either remove all the occurrences of single or multiple items from a list

suresh-prakash: Are you proposing we convert each operator to a one of message?

skjindal93: > Are you proposing we convert each operator to a one of message?

We can. Could be an overkill. What do you think?

suresh-prakash: > Could be an overkill.

Agree. Keeping it as an operator makes the mapping easier. Whether to set the `LiteralConstant` etc. would be implied by the operator itself.

> rename REMOVE_ALL_FROM_LIST to REMOVE_FROM_LIST, and may be have an UNSET option too

Sure, will add.

skjindal93 (approved):

PR #212, opened 2022-11-23
fix(eqs): missing tenant id filter in bulkUpdateEntityArrayAttribute

skjindal93:


avinashkolluru (approved):

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -367,6 +367,12 @@ public void update(EntityUpdateRequest request, StreamObserver<ResultSetChunk> r
 
   private void doUpdate(RequestContext requestContext, EntityUpdateRequest request)
       throws Exception {
+    Optional<String> maybeTenantId = requestContext.getTenantId();
+    if (maybeTenantId.isEmpty()) {
+      return;
```
kotharironak, line 372: Will we throw an exception eventually as this one is private?

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -564,6 +570,12 @@ private List<Document> getProjectedDocuments(
   private void doBulkUpdate(
       RequestContext requestContext, String entityType, Map<String, EntityUpdateInfo> entitiesMap)
       throws Exception {
+    Optional<String> maybeTenantId = requestContext.getTenantId();
+    if (maybeTenantId.isEmpty()) {
+      return;
```
kotharironak, line 575: Will we throw an exception eventually to end API caller?

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
skjindal93: The caller already checks the presence of tenant id in request context, and returns an error

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
skjindal93: The caller already checks the presence of tenant id in request context, and returns an error

suresh-prakash: In that case, could have simplified to
```
String tenantId = requestContext.getTenantId().orElseThrow();
```

Just by looking at the `doBulkUpdate()` method, it feels like it is a no-op when the tenantId is missing. Throwing exception would improve the readability like `tenantId` is expected to be present otherwise the operation would fail.

kotharironak: If a caller is making sure, we don't need these additional checks.

PR #216, opened 2022-12-21
Return updated entity ids in bulk update response

suresh-prakash:


entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -827,8 +827,10 @@ private void doBulkUpdate(
 
       if (keys.isEmpty()) {
         // Nothing to update
-        LOG.info("No entity found with filter {} for updating", update.getFilter());
-        continue;
+        throw new IllegalArgumentException(
```
aaron-steinfeld, line 830: Hmm - I know I reported this, but I'm less sure we actually want to treat this as an error now that I see this. 

I think there's a valid use case to want to update based on some criteria, and not have it error out if it doesn't hit anything. That said, if we go that way, I'd expect it to return the updated count if it doesn't already (so the caller can decide if zero matches is an error to it or not) - and I'm guessing by the void return on this method that it doesn't do that.

So I think returning the update count would be better than an error, but if that's not feasible I guess we can stick with the error for now.

suresh-prakash: Sure, Aaron. That makes sense. Updated to return the updated ids.

entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/response.proto
```diff
@@ -41,4 +41,9 @@ message DeleteEntitiesResponse {
 }
 
 message BulkUpdateAllMatchingFilterResponse {
+  repeated UpdateSummary summaries = 1;
+}
+
+message UpdateSummary {
+  repeated string ids = 1;
```
aaron-steinfeld, line 48: could we add types here too

entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/response.proto
```diff
@@ -41,4 +41,13 @@ message DeleteEntitiesResponse {
...
 message BulkUpdateAllMatchingFilterResponse {
+  repeated UpdateSummary summaries = 1;
+}
+
+message UpdateSummary {
+  repeated UpdatedEntity responses = 1;
+}
+
+message UpdatedEntity {
+  string id = 1;
```
aaron-steinfeld, line 52: sorry poorly worded my other comment, I meant entity types on the response (since they can be mixed, right?). I don't think we need the double-repeated:
```suggestion
  repeated UpdatedEntity updated_entities = 1;
}

message UpdatedEntity {
  string type = 1;
  string id = 2;
```

suresh-prakash: > since they can be mixed, right?

Nope. All the updates should be of the same entity type. Whereas different update operations are supported based on different filters.

We need 2 repeated because the top level is for the number of update operations and the second level is for the actual entity ids that got updated. Since each update operation could update multiple entities, we need a repeated there also.
```
message BulkUpdateAllMatchingFilterRequest {
  string entity_type = 1;
  repeated Update updates = 2;
}

message Update {
  Filter filter = 1;
  repeated AttributeUpdateOperation operations = 2;
}
```

aaron-steinfeld: got it.

aaron-steinfeld: got it. seems fine to me then

entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/response.proto
```diff
@@ -41,4 +41,13 @@ message DeleteEntitiesResponse {
 }
 
 message BulkUpdateAllMatchingFilterResponse {
+  repeated UpdateSummary summaries = 1;
+}
+
+message UpdateSummary {
+  repeated UpdatedEntity responses = 1;
```
aaron-steinfeld, line 48: nit:
```suggestion
  repeated UpdatedEntity updated_entities = 1;
```

aaron-steinfeld (dismissed):

aaron-steinfeld (approved):

PR #217, opened 2022-12-26
Eng 25032-add support for list of maps for Entity Query Service

bhubam:
## Description
Add support for list of maps for Entity Query Service

### DIFF
--- a/entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/value.proto
+++ b/entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/value.proto
@@ -22,6 +22,14 @@ enum ValueType {
   BOOLEAN_ARRAY = 14;
   // assumes that key is always string
   STRING_MAP = 15;
+  LONG_MAP = 16;
+  INT_MAP = 17;
+  FLOAT_MAP = 18;
+  DOUBLE_MAP = 19;
+  BYTES_MAP = 20;
+  BOOLEAN_MAP = 21;
+  VALUE_MAP = 22;
+  VALUE_LIST = 23;
 }
 
 message Value {
@@ -43,4 +51,12 @@ message Value {
   repeated bytes bytes_array = 21;
   repeated bool boolean_array = 22;
   map<string, string> string_map = 23;
+  map<string, int64> long_map = 24;
+  map<string, int32> int_map = 25;
+  map<string, float> float_map = 26;
+  map<string, double> double_map = 27;
+  map<string, bool> boolean_map = 28;
+  map<string, bytes> bytes_map = 29;
+  map<string, Value> value_map = 30;
+  repeated Value value_list = 31;
 }
--- a/entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/value.proto
+++ b/entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/value.proto
@@ -22,7 +22,7 @@ enum ValueType {
   BOOLEAN_ARRAY = 14;
   STRING_MAP = 15;
   VALUE_MAP = 16;
-  VALUE_LIST = 17;
+  VALUE_ARRAY = 17;
 }
 
 message Value {
--- a/entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/value.proto
+++ b/entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/value.proto
@@ -22,14 +22,8 @@ enum ValueType {
   BOOLEAN_ARRAY = 14;
   // assumes that key is always string
   STRING_MAP = 15;
-  LONG_MAP = 16;
-  INT_MAP = 17;
-  FLOAT_MAP = 18;
-  DOUBLE_MAP = 19;
-  BYTES_MAP = 20;
-  BOOLEAN_MAP = 21;
-  VALUE_MAP = 22;
-  VALUE_LIST = 23;
+  VALUE_MAP = 16;
+  VALUE_LIST = 17;
 }
 
 message Value {
@@ -51,12 +45,6 @@ message Value {
   repeated bytes bytes_array = 21;
   repeated bool boolean_array = 22;
   map<string, string> string_map = 23;
-  map<string, int64> long_map = 24;
-  map<string, int32> int_map = 25;
-  map<string, float> float_map = 26;
-  map<string, double> double_map = 27;
-  map<string, bool> boolean_map = 28;
-  map<string, bytes> bytes_map = 29;
-  map<string, Value> value_map = 30;
-  repeated Value value_list = 31;
+  map<string, Value> value_map = 24;
+  repeated Value value_list = 25;
 }
--- a/entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/value.proto
+++ b/entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/value.proto
@@ -20,7 +20,6 @@ enum ValueType {
   DOUBLE_ARRAY = 12;
   BYTES_ARRAY = 13;
   BOOLEAN_ARRAY = 14;
-  // assumes that key is always string
   STRING_MAP = 15;
   VALUE_MAP = 16;
   VALUE_LIST = 17;
@@ -46,5 +45,5 @@ message Value {
   repeated bool boolean_array = 22;
   map<string, string> string_map = 23;
   map<string, Value> value_map = 24;
-  repeated Value value_list = 25;
+  repeated Value value_array = 25;
 }
--- a/entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/response/getter/ArrayGetter.java
+++ b/entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/response/getter/ArrayGetter.java
@@ -18,27 +18,16 @@ import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
 import org.hypertrace.entity.query.service.converter.ConversionException;
-import org.hypertrace.entity.query.service.converter.ValueHelper;
 import org.hypertrace.entity.query.service.v1.Value;
 import org.hypertrace.entity.query.service.v1.ValueType;
 
 @Singleton
 public class ArrayGetter implements ValueGetter {
-  private final ValueGetter nestedValueGetter;
-  private final ValueGetter directValueGetter;
   private final List<ValueGetter> rootGetters;
-  private final ValueHelper valueHelper;
 
   @Inject
-  public ArrayGetter(
-      @Named("nested_value") final ValueGetter nestedValueGetter,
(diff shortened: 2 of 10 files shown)


entity-service-api/src/main/proto/org/hypertrace/entity/query/service/v1/value.proto
```diff
@@ -43,4 +45,6 @@ message Value {
   repeated bytes bytes_array = 21;
   repeated bool boolean_array = 22;
   map<string, string> string_map = 23;
+  map<string, Value> value_map = 24;
+  repeated Value value_list = 25;
```
suresh-prakash, line 49: Nit:
```suggestion
  repeated Value value_array = 25;
```
to be consistent with the other field names.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/response/getter/ListValueGetter.java
```diff
@@ -8,16 +8,21 @@
...
   private final ValueGetter arrayGetter;
+  private final List<ValueGetter> rootGetters;
 
   @Inject
-  public ListValueGetter(@Named("array") final ValueGetter arrayGetter) {
+  public ListValueGetter(
+      @Named("array") final ValueGetter arrayGetter,
+      @Named("root_getters") final List<ValueGetter> rootGetters) {
     this.arrayGetter = arrayGetter;
+    this.rootGetters = rootGetters;
```
suresh-prakash, line 25: Wondering if we could include the `arrayGetter` in the `rootGetters`!
Could you explore this?

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/response/getter/MapValueGetter.java
```diff
@@ -56,16 +61,26 @@ public Value getValue(final JsonNode jsonNode) throws ConversionException {
 
       final String key = entry.getKey();
       final JsonNode node = entry.getValue();
-
-      if (!nestedValueGetter.matches(node)) {
+      boolean valueSet = false;
+      if (nestedValueGetter.matches(node)) {
```
suresh-prakash, line 65: Could this be combined with the for-loop below (since nestedValueGetter is already a part of the rootGetters).

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/response/getter/MapValueGetter.java
```diff
@@ -56,16 +61,26 @@ public Value getValue(final JsonNode jsonNode) throws ConversionException {
...
+        final Value value = nestedValueGetter.getValue(node);
+        Object obj = valueOneOfAccessor.access(value, value.getValueType());
+        valueBuilder.putStringMap(key, obj.toString());
+        continue;
+      }
+      for (ValueGetter getter : rootGetters) {
+        if (getter.matches(node)) {
+          final Value value = this.getValue(node);
```
suresh-prakash, line 73: Shouldn't this be getter.getValue(node) instead?

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/response/getter/MapValueGetter.java
```diff
@@ -56,16 +61,26 @@ public Value getValue(final JsonNode jsonNode) throws ConversionException {
...
+          valueBuilder.setValueType(VALUE_MAP);
```
suresh-prakash, line 74: Maybe, we need to check the value type here?
If the value type is `STRING`, then we should set this to `STRING_MAP`.
If the value type is anything else, then we should set this to `VALUE_MAP`.

suresh-prakash:
Also, can you add 2 more for unit testing?
1. String lists
2. Value lists
3. Map of lists
4. List of maps

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/response/getter/ListValueGetter.java
bhubam: arrayGetter is always called in ListValueGetter.  ListValueGetter firstly always checks if node matches with `VALUES_LIST_KEY` and then we check if the node is an array, so I think we don't need to add arrayGetters in rootGetters as, array will always come inside the `valuesList` and arrayGetters.match() won't match anything in the ValueMapper.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/response/getter/MapValueGetter.java
bhubam: yes, my mistake

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/response/getter/ListValueGetter.java
suresh-prakash: Got it! Thanks for checking this.

entity-service-impl/src/test/java/org/hypertrace/entity/query/service/converter/response/DocumentConverterTest.java
```diff
@@ -82,6 +88,35 @@ void testConvert() throws IOException, ConversionException {
...
+            .addColumn(
+                Value.newBuilder()
+                    .setValueType(VALUE_MAP)
+                    .putValueMap(
+                        "countries",
+                        Value.newBuilder()
+                            .setValueType(STRING_ARRAY)
+                            .addAllStringArray(List.of("india", "australia"))
+                            .build())
+                    .build())
```
suresh-prakash, line 119: Also, add a case for array of maps.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/response/getter/ArrayGetter.java
```diff
@@ -43,53 +54,110 @@ public Value getValue(final JsonNode jsonNode) throws ConversionException {
...
 
-      case BOOLEAN_ARRAY:
+      case BOOL:
         values.stream().map(Value::getBoolean).forEach(valueBuilder::addBooleanArray);
+        valueBuilder.setValueType(BOOLEAN_ARRAY);
+        break;
+
+      case INT_ARRAY:
+        values.stream().map(Value::getIntArrayList).forEach(valueBuilder::addAllIntArray);
+        valueBuilder.setValueType(VALUE_ARRAY);
```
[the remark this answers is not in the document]
suresh-prakash: Same everywhere.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/response/getter/ListValueGetter.java
```diff
@@ -8,16 +8,21 @@
...
 import org.hypertrace.entity.query.service.v1.Value;
 
 @Singleton
 public class ListValueGetter implements ValueGetter {
   private final ValueGetter arrayGetter;
+  private final List<ValueGetter> rootGetters;
 
   @Inject
-  public ListValueGetter(@Named("array") final ValueGetter arrayGetter) {
+  public ListValueGetter(
```
suresh-prakash, line 21: Can we revert all the changes in this file for a clean git history?

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/response/getter/ArrayGetter.java
```diff
@@ -18,15 +26,18 @@
 public class ArrayGetter implements ValueGetter {
   private final ValueGetter nestedValueGetter;
   private final ValueGetter directValueGetter;
+  private final List<ValueGetter> rootGetters;
   private final ValueHelper valueHelper;
 
   @Inject
   public ArrayGetter(
       @Named("nested_value") final ValueGetter nestedValueGetter,
       @Named("direct_value") final ValueGetter directValueGetter,
```
suresh-prakash, line 35: The first 2 constructor arguments seem to be redundant. Can we get rid of them?

entity-service-impl/src/test/resources/response/nested_document.json
```diff
@@ -1,4 +1,94 @@
 {
+  "countries": {
+    "valueList": {
+      "values": [ "india", "australia" ]
```
suresh-prakash, line 4: Please reference the format:
```
{
  "region":
  {
    "valueMap":
    {
      "values":
      {
        "countries":
        {
          "valueList":
          {
            "values": [{"value": {"string": "india"}},{"value": {"string": "australia"}}]
          }
        }
      }
    }
  }
},
{
  "timestamp_1":
  {
    "valueList":
    {
      "values": [{"valueMap": {"values": {"seconds": {"value": {"long": 10}}}}}]
    }
  }
}
```

entity-service-impl/src/test/resources/response/nested_document.json
```diff
@@ -1,4 +1,94 @@
...
+        }
+        ]
+      }
+  },
+  "region": {
+    "valueMap": {
+      "values": {
+        "countries": {
+          "valueList" : {
+            "values": [ "india", "australia" ]
```
suresh-prakash, line 54: Fix format.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/response/getter/ArrayGetter.java
```diff
@@ -43,53 +43,74 @@ public Value getValue(final JsonNode jsonNode) throws ConversionException {
...
+
+      case INT_ARRAY:
+      case LONG_ARRAY:
+      case STRING_ARRAY:
+      case FLOAT_ARRAY:
+      case DOUBLE_ARRAY:
+      case BYTES_ARRAY:
+      case BOOLEAN_ARRAY:
+      case STRING_MAP:
+      case VALUE_MAP:
```
avinashkolluru, line 110: Dont have the complete context but this doesn't seem to be correct. Why is ValueArray being set in the case of VALUE_MAP?

suresh-prakash: Basically, we are checking the "containing type".
If the containing type is anything other than a primitive, the current type is a nested array (represented as "VALUE_ARRAY").

@bhubam For clarity, could you rename the variable to "containingType" in line 63?

entity-service-impl/src/test/resources/response/nested_document.json
```diff
@@ -1,4 +1,101 @@
...
+            "values": {
+              "nanos": {
+                "value": {
+                  "long": 20
+                }
+              }
+            }
+          }
+        }
+        ]
```
suresh-prakash, line 48: Please reformat the entire file once.
`spotlessApply` doesn't work on `json` files.

suresh-prakash (approved):

PR #220, opened 2023-01-19
Validate LHS and RHS type combinations and LHS to operator combinations for bulk update

suresh-prakash:


entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/UpdateConverter.java
```diff
@@ -89,31 +79,39 @@ public class UpdateConverter implements Converter<AttributeUpdateOperation, SubD
...
+        entityAttributeMapping.getAttributeKind(context, id);
 
-    if (pathOptional.isEmpty()) {
+    if (attributeKindOptional.isEmpty()) {
       throw new ConversionException(String.format("Cannot update non-attribute value for %s", id));
     }
 
-    final String subDocPath = pathOptional.orElseThrow();
-    final UpdateOperator operator = metadata.operator();
+    final AttributeKind attributeKind = attributeKindOptional.orElseThrow();
```
aaron-steinfeld, line 97: nit - duplicative of 93. Instead remove that if condition, 

```suggestion
    final AttributeKind attributeKind = attributeKindOptional.orElseThrow(() => some error);
```

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/UpdateConverter.java
```diff
@@ -89,31 +79,39 @@ public class UpdateConverter implements Converter<AttributeUpdateOperation, SubD
...
+    validateOperator(attributeKind, operation.getOperator());
+
     final Value value = operation.getValue().getValue();
+    // Validate if the LHS and the RHS are of same type
+    validateDataType(attributeKind, operation.getOperator(), value.getValueType());
+
+    final Optional<String> pathOptional =
+        entityAttributeMapping.getDocStorePathByAttributeId(context, id);
+
+    final String subDocPath = pathOptional.orElseThrow();
```
aaron-steinfeld, line 108: likewise here.

entity-service-attribute-translator/src/main/java/org/hypertrace/entity/attribute/translator/EntityAttributeMapping.java
```diff
@@ -22,6 +28,9 @@ public class EntityAttributeMapping {
   private static final String SCOPE_PATH = "scope";
   private static final String ATTRIBUTE_PATH = "attribute";
   private static final String NAME_PATH = "name";
+  private static final Set<AttributeKind> MULTI_VALUED_ATTRIBUTE_KINDS =
+      Set.of(
+          TYPE_STRING_ARRAY, TYPE_INT64_ARRAY, TYPE_DOUBLE_ARRAY, TYPE_BOOL_ARRAY, TYPE_STRING_MAP);
```
aaron-steinfeld, line 33: Maps aren't really multivalued in the same ways as arrays. They support different operators, and for example later you write

```java
    if (entityAttributeMapping.isMultiValued(attributeKind)) {
      suffixedSubDocPath = DOT_JOINER.join(subDocPath, VALUE_LIST_KEY, VALUES_KEY);
```

That would fail for a map, no?

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/UpdateConverter.java
```diff
@@ -89,31 +79,39 @@ public class UpdateConverter implements Converter<AttributeUpdateOperation, SubD
   public SubDocumentUpdate convert(
       final AttributeUpdateOperation operation, final RequestContext context)
       throws ConversionException {
-    final AttributeOperationMetadata metadata = OPERATOR_MAP.get(operation.getOperator());
+    final UpdateOperator operator = OPERATOR_MAP.get(operation.getOperator());
 
-    if (metadata == null) {
+    if (operator == null) {
       throw new ConversionException(
```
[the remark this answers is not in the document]
suresh-prakash: Just trying to understand. 
This is not a validator class. This is a part of the converters. Do we still want to through Status exceptions from here?

Also, what if this converter class was being called by many clients (a few of them catching the custom `ConversionException` for handling the exception scenarios differently and others rethrowing)?
Throwing StatusRuntimeException may make the exception part of this method's signature implicit (and hence the clients may forget to catch runtime exceptions).

Moreover, even if we were throwing Status exception from here, it would be caught in `org.hypertrace.entity.query.service.EntityQueryServiceImpl.bulkUpdateAllMatchingFilter`, which would then wrap this in an `INTERNAL` status exception. So, the clients would anyways be receiving an `INTERNAL` error despite with through Status exception or others from here.

> grpc status with trailers

Sure, I'll add the trailers to the original GRPC handler method.

entity-service-attribute-translator/src/main/java/org/hypertrace/entity/attribute/translator/EntityAttributeMapping.java
suresh-prakash: Good catch, Aaron. I'll remove the map from here.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/UpdateConverter.java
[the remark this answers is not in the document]
aaron-steinfeld: Yeah, fair point - it depends on how it's used. The main goal is that anything returned from the service returns a meaningful error rather than an "internal - unknown" type error. For this, if I'm reading it correctly - doesn't this mean the caller sent an operator that was not supported? So that seems useful to propagate back as an invalid argument.

There may be bigger refactors needed to make that work cleanly and will let you judge if it's in scope or not, but that's the high level point - that the vast majority of exceptions we throw are intended to propagate back, and should be constructed in a form to support that. If it is intended to be caught by the caller, then - dare I say it? - perhaps a checked exception is a better fit (which I now see you're using here). So who is looking for that checked exception, and what are they doing with it?

aaron-steinfeld (approved):

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/UpdateConverter.java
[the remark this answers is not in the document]
suresh-prakash: > doesn't this mean the caller sent an operator that was not supported?

That's right! If we had to make this a part of the request validator, it would have to query attribute service. Not sure if it is a good idea to make the validator talk to external services. Moreover, on-the-fly validation designates validating the input to different classes (each one will be responsible for validating and processing). If we had to do that in the validator class, we would have to get the object that could be deeply nested and perform the validation.


I've added a version to capture it and convert it into an INVALID_ARGUMENT. Please take a look and let me know if there's any better way to handle it. Note: I have not done it for all the code paths in bulk update. Did only for the paths I touched in this PR.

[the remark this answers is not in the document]
suresh-prakash: Basically, the idea is to throw Status exception (but as a checked exception than as a runtime exception).
Any checked exception would be caught by the service implementation method, logged, and notified to the clients.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -808,8 +809,19 @@ public void bulkUpdateAllMatchingFilter(
       final BulkUpdateAllMatchingFilterResponse response = doBulkUpdate(request, requestContext);
       responseObserver.onNext(response);
       responseObserver.onCompleted();
+    } catch (final StatusException e) {
+      LOG.warn(
+          String.format(
+              "Error while bulk updating entities with request %s for %s",
+              request, tenantIdOptional.orElse("")),
```
aaron-steinfeld, line 816: instead of logging tenant id, log the context.

Also can use slf4j log string template instead of explicit string format.

aaron-steinfeld (commented):
seems like a reasonable approach

aaron-steinfeld (approved):

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
suresh-prakash: Oh! That's actually smart. 🙂

suresh-prakash: Continuing with the merge since it's a nit. Will probably piggy back the changes in some subsequent changes.

PR #221, opened 2023-01-20
chore | adding rate limiter to entity service

aman-bansal:


aaron-steinfeld (commented):
I think I'm missing context here - why would we want to rate limit entity writes? I could see why we might want to from the _client_ side, but by doing it here we're just asking to lose data.

PR #229, opened 2023-03-23
Reverted #167 to remove the queryAggregationEnabled flag

sanket-mundra:
Removed the queryAggregationEnabled flag as the aggregate() is supported for both mongo and postgres now.


entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -183,11 +172,9 @@ public EntityQueryServiceImpl(
         entityCounterMetricSender,
         new EntityFetcher(entitiesCollection, DOCUMENT_PARSER),
         chunkSize,
-        queryAggregationEnabled,
         maxEntitiesToDelete);
   }
 
-  @VisibleForTesting
```
sanket-mundra, line 190: The above constructor itself is calling this constructor: https://github.com/hypertrace/entity-service/pull/229/files#diff-0aad350d7986d390c6ff43b5f9074b244ea7b1a7e1baef820a4eb4b70a58f5b3L178

Apart from the test files, it is being called at this place only.
image

suresh-prakash (approved):

suresh-prakash:
Please make sure to remove the irrelevant sections from the PR description

suresh-prakash (approved):

PR #234, opened 2023-04-24
fix: Get the ValueType from db instead of RHS

sanket-mundra:
## Description

Get the ValueType from db instead of rhs value

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -54,7 +54,7 @@ jobs:
         uses: hypertrace/github-actions/gradle@main
         with: 
           args: copyAllReports --output-dir=/tmp/test-reports
-        if: always()
+        if: failure()
             
       - name: Archive test reports
         uses: actions/upload-artifact@v1
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -61,7 +61,7 @@ jobs:
         with:
           name: test-reports
           path: /tmp/test-reports
-        if: always()
+        if: failure()
      
       - name: Publish Unit Test Results
         uses: docker://ghcr.io/enricomi/publish-unit-test-result-action:v2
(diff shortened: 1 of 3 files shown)


.github/workflows/pr-test.yml
```diff
@@ -54,6 +54,7 @@ jobs:
         uses: hypertrace/github-actions/gradle@main
         with: 
           args: copyAllReports --output-dir=/tmp/test-reports
+        if: always()
```
suresh-prakash, line 57: Why is this change required?

suresh-prakash: Can you confirm with @shaishaw if that is not an issue?

shaishaw: Is the intention is to execute the step "copy test reports" on all case ? then it's fine to keep always flag.

sanket-mundra: What should be the ideal scenario? We want to see the result of integration test only in case of failure? Also, this is how it was in other repos as well so, I did that: https://github.com/hypertrace/attribute-service/blob/main/.github/workflows/pr-test.yml#L40

shaishaw: If you want it to run only incase of failure, you can try with other options.

you can try with if: failure()

aaron-steinfeld: We want test results whether success or failure of the tests. More accurate would be to look for the test task to succeed or fail (e.g. `success() || failure()`) to avoid upload in cases where the task isn't reached/cancelled, but we usually just use always() as a shorthand since it's more readable and even in the cancel case a failed upload doesn't really matter.

sanket-mundra: Updated it to `if: failure()`

sanket-mundra: We will need to update the same for `Archive test reports` job as well as the report will be copied only in case of failure and only then we need to archive it:

aaron-steinfeld: We typically upload on success too - you can use the other repos you were looking at for reference.

sanket-mundra: I other repos, we are simply using `if: always()`. So, reverting back to `if: always()`

suresh-prakash (approved):

PR #235, opened 2023-05-01
fix: timestamp filters should be mapped to longs

aaron-steinfeld:
## Description

Timestamps are their own type, but all writers store them as longs. When filtering, we need to map timestamp attributes back to that storage format.


entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/filter/PrimitiveFilteringExpressionConverter.java
```diff
@@ -61,7 +62,9 @@ public class PrimitiveFilteringExpressionConverter extends FilteringExpressionCo
           entry(TYPE_INT64_ARRAY, LONG_ARRAY),
           entry(TYPE_DOUBLE_ARRAY, DOUBLE_ARRAY),
           entry(TYPE_BOOL_ARRAY, BOOLEAN_ARRAY),
-          entry(TYPE_STRING_MAP, STRING_MAP));
+          entry(TYPE_STRING_MAP, STRING_MAP),
+          entry(TYPE_TIMESTAMP, LONG) // Timestamps are stored as longs
```
aaron-steinfeld, line 66: There may be more to do here - this is confusing - but I tested this e2e and followed all code paths on the write side. All current writers persist a timestamp as a long.

Fundamentally, there's a bigger issue here. The type comes in 3 different spots.

- the schema
- the value writer
- the value reader

 https://github.com/hypertrace/entity-service/pull/234 aligned the schema and reader, but writer is assumed to match the schema which is not always the case. Ideally, the schema should be used to transform any written values (and done in a compatible way).

tim-mwangi (approved):

PR #238, opened 2023-05-30
chore | preserving null and empty values for multi valued attribute

aman-bansal:


skjindal93 (approved):

suresh-prakash (approved):

entity-service/src/integrationTest/java/org/hypertrace/entity/service/service/EntityQueryServiceTest.java
```diff
@@ -981,6 +997,194 @@ void testExecuteWithExistsFilter() {
     assertTrue(list.get(1).getResultSetMetadata().getColumnMetadataCount() > 0);
   }
 
+  @Test
```
kotharironak, line 1000: Does the test run against both the DB? Or Did we try manually against Postgres?

kotharironak: @aman Yes, the test runs agains only MongoDB. Can we grep the translated docStore query and see if its compatible? or Can you paste the translated query here?

PR #241, opened 2023-06-20
chore: update libs to fix vulnerabilities

sanket-mundra:
## Description
Please include a summary of the change, motivation and context.

### Testing

### Checklist:

### Documentation


entity-data-service-rx-client/build.gradle.kts
```diff
@@ -13,11 +13,11 @@ dependencies {
...
 
-  implementation("org.hypertrace.core.grpcutils:grpc-client-utils:0.11.2")
-  implementation("org.hypertrace.core.grpcutils:grpc-client-rx-utils:0.11.2")
+  implementation("org.hypertrace.core.grpcutils:grpc-client-utils:0.12.0")
+  implementation("org.hypertrace.core.grpcutils:grpc-client-rx-utils:0.12.0")
 
   implementation("org.slf4j:slf4j-api:1.7.30")
-  implementation("com.google.guava:guava:31.1-jre")
+  implementation("com.google.guava:guava:32.0.1-jre")
   implementation("org.slf4j:slf4j-api:1.7.30")
```
aaron-steinfeld, line 21: nit - pre-existing, but please remove dupe declaration (it's right above on 19 too)

entity-data-service-rx-client/build.gradle.kts
```diff
@@ -13,11 +13,11 @@ dependencies {
...
   annotationProcessor("org.projectlombok:lombok:1.18.18")
   compileOnly("org.projectlombok:lombok:1.18.18")
```
aaron-steinfeld, line 23: Make sure all grpc libs are upgraded. I see a one on line 28 in the file that hasn't been updated.

entity-service/build.gradle.kts
```diff
@@ -22,7 +22,7 @@ tasks.integrationTest {
 dependencies {
   implementation(project(":entity-service-factory"))
 
-  implementation("org.hypertrace.core.serviceframework:platform-grpc-service-framework:0.1.49")
+  implementation("org.hypertrace.core.serviceframework:platform-grpc-service-framework:0.1.52")
   implementation("org.hypertrace.core.documentstore:document-store:0.7.29")
 
   runtimeOnly("io.grpc:grpc-netty:1.50.0")
```
aaron-steinfeld, line 28: especially this one - mismatching the grpc runtime (netty) with the grpc libs will cause a runtime failure.

aaron-steinfeld:
Also suggest moving the scan over either before this PR or as part of it, since right now no vuln scan is running to validate your changes. Example PR: https://github.com/hypertrace/java-grpc-utils/pull/46

owasp-suppressions.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.3.xsd">
+    <suppress>
+        <notes><![CDATA[
+   file name: avro-1.9.2.jar
```
aaron-steinfeld, line 5: Please include an explanation of why we're suppressing this vulnerability (e.g. "According to https://nvd.nist.gov/vuln/detail/cve-2021-43045 , CVE only impacts .NET SDK")

sanket-mundra:
What should we do for vulnerability in `snappy-java-1.1.8.4.jar`?

entity-service-api/build.gradle.kts
```diff
@@ -61,7 +61,12 @@ sourceSets {
 }
 
 dependencies {
-  api("io.grpc:grpc-protobuf:1.50.0")
-  api("io.grpc:grpc-stub:1.50.0")
+  api("io.grpc:grpc-protobuf:1.56.0")
```
aaron-steinfeld, line 64: Ref https://github.com/hypertrace/gateway-service/pull/165#discussion_r1237267427

entity-service-change-event-generator/build.gradle.kts
```diff
@@ -27,7 +27,7 @@ dependencies {
     }
   }
 
-  runtimeOnly("io.confluent:kafka-protobuf-serializer:6.0.1")
+  runtimeOnly("io.confluent:kafka-protobuf-serializer:7.4.0")
```
aaron-steinfeld, line 30: Here, we should use the kafka bom. That should look something like

```
runtimeOnly("io.confluent:kafka-protobuf-serializer")

implementation(platform("org.hypertrace.core.kafkastreams.framework:kafka-bom:0.2.10")
```

Let me try to update a repo to get an example.

aaron-steinfeld: > > What should we do for vulnerability in `snappy-java-1.1.8.4.jar`?

Also the kafka bom will fix this

owasp-suppressions.xml
```diff
@@ -39,6 +39,14 @@
         <packageUrl regex="true">^pkg:maven/com\.squareup\.wire/wire\-schema\-jvm@.*$</packageUrl>
         <cpe>cpe:/a:wire:wire</cpe>
     </suppress>
+    <suppress>
+        <notes><![CDATA[
+   file name: protobuf-java-util-3.19.4.jar
+   mismatching with kafka-protobuf-serializer
```
aaron-steinfeld, line 45: this package and this cpe seem to match (the comment mentions kafka which doesn't appear in the actual suppression data). If we're using a protobuf that old, it likely _does_ have vulns and should be updated (try 3.21.7)

sanket-mundra: This is the dependency check result that I am getting without this suppression:
image
According to this the `protobuf-java-util-3.19.4.jar` dependency is coming from `kafka-bom (kafka-protobuf-serializer )` itself. What should we do here?

aaron-steinfeld: ok so it's a real issue, and the vulnerable lib is a dependency of kafka. We can either update the version here, or update it in our bom and then include the updated bom here. If it's not used other than via kafka (i.e. not directly), I'd say we should do the bom else the local constraint. We often use this lib, so it's probably already in the repo.

aaron-steinfeld: https://github.com/hypertrace/kafka-streams-framework/pull/64

sanket-mundra:
On local, `dependencyCheckAggregate` is passing with 0 vulns but here it's failing. Not sure what's going wrong.

aaron-steinfeld:
> On local, `dependencyCheckAggregate` is passing with 0 vulns but here it's failing. Not sure what's going wrong.

Caching. If you  do a `dependencyCheckPurge` for example first, it'll rebuild the DB to get the latest and you should see it.

aaron-steinfeld:
Fixed in https://github.com/hypertrace/java-grpc-utils/pull/47

.github/workflows/pr-test.yml
```diff
@@ -57,15 +44,21 @@ jobs:
...
      
       - name: Publish Unit Test Results
-        uses: docker://ghcr.io/enricomi/publish-unit-test-result-action:v2
+        uses: EnricoMi/publish-unit-test-result-action@v2
         if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           files: ./**/build/test-results/**/*.xml
+
+  dependency-check:
```
ravisingal, line 60: this should be part of `pr-build.yaml`

.github/workflows/pr-build.yml
```diff
@@ -9,28 +9,16 @@ on:
```
ravisingal, line 9: add a cron also:
```
on:
    schedule:
    - cron: 10 0 * * 1
```

ravisingal:
@sanket-mundra use latest github action for scanning image:
```
      - name: Run Trivy vulnerability scanner for kafka-zookeeper image
        uses: hypertrace/github-actions/trivy-image-scan@main
        with:
          image: hypertrace/kafka-zookeeper
          output-mode: github
```

.github/workflows/pr-test.yml
sanket-mundra: Followed this to add depencycheck scan: https://github.com/hypertrace/java-grpc-utils/pull/46 and here its part of `pr-test.yml` 😅

aaron-steinfeld: Two competing concerns here. The test job and build job have different triggers based on how they use secrets. We want to put all jobs under the `pull_request` trigger rather than `pull_request_target` unless there's a reason not to, so we'll need to take a look at how we can best arrange these so that we can get everything needed scheduled, but not over-expose on the the pull_request_target trigger. Will put something up later this morning.

aaron-steinfeld: OK updated attribute service as an example - we don't really need that separation of pull_request and pull_request_target any more, so we can merge the two into a single workflow that gets scheduled with a few changes. You may want to do that in a separate PR though, up to you.

https://github.com/hypertrace/attribute-service/pull/170/files

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,124 @@
...
+    steps:
+      - name: Dependency Check
+        uses: hypertrace/github-actions/dependency-check@main
+
+  validate-protos:
+    runs-on: ubuntu-22.04
+    container:
+      image: bufbuild/buf:0.56.0
+      credentials:
+        username: ${{ secrets.DOCKERHUB_READ_USER }}
```
aaron-steinfeld, line 100: Shouldn't use creds here. This is using the old buf version, we should upgrade to the GA buf (1.0+).

That'll look like 
```yaml
      - name: Setup buf
        uses: bufbuild/buf-setup-action@v1
        with:
          github_token: ${{ github.token }}
      - name: Lint protos
        uses: bufbuild/buf-lint-action@v1
      - name: Check for breaking changes
        uses: bufbuild/buf-breaking-action@v1
        with:
          against: '.git#branch=origin/main'
```

It also means the buf config needs to be updated though so my suggestion for now would be to just remove the creds from here - let it run with anonymous pulls and we can come through and upgrade it after.

aaron-steinfeld (approved):

PR #248, opened 2023-09-06
chore: add creation source entity type v2 field

SrikarMannepalli:
## Description
This PR introduces a creation source entity type v2 field.


entity-service-api/src/main/proto/org/hypertrace/entity/type/service/v2/entity_type_service.proto
```diff
@@ -10,6 +10,7 @@ message EntityType {
   string name_attribute_key = 4;
   string timestamp_attribute_key = 5;
   repeated EntityFormationCondition required_conditions = 6;
+  string create_source = 7;
```
aaron-steinfeld, line 13: would make this optional (pending discussion about alternative approaches)

PR #254, opened 2023-10-17
Document store config overhaul

suresh-prakash:


helm/templates/deployment.yaml
```diff
@@ -83,6 +83,20 @@ spec:
               value: "/var/{{ .Chart.Name }}/log/log4j2.properties"
             - name: JAVA_OPTS
               value: {{ .Values.javaOpts | quote }}
+            {{- if .Values.database and .Values.database.mongo.authEnabled }}
```
suresh-prakash, line 86:
```suggestion
            {{- if and .Values.database .Values.database.mongo.authEnabled }}
```

helm/templates/entity-service-config.yaml
```diff
@@ -7,20 +7,27 @@ metadata:
...
+        appName = {{ .Values.service.name }}
         {{ $dst }} {
-          {{- range $key, $value := (index .Values "entityServiceConfig" (printf "%s" $dst)) }}
+          {{- range $key, $value := (index .Values "database" (printf "%s" $dst)) }}
           {{- if $value }}
-            {{- if eq $key "password" }}
+            {{- if hasPrefix "${?" (printf "%s" $value) }}
           {{ $key }} = {{ $value }}
             {{- else }}
+              {{- if eq $key "endpoints" }}
```
aaron-steinfeld, line 21: do we need to special case this? Can we we toJson everything (I'm assuming a standard string would just quote but I'm not sure)

helm/values.yaml
```diff
@@ -92,10 +92,6 @@ serviceSelectorLabels:
 entityServiceConfig:
   name: entity-service-config
   publishChangeEvents: false
-  dataStoreType: "mongo"
```
aaron-steinfeld, line 95: So previously the values defaulted it to mongo - can we provide that equivalent config in the new format?

aaron-steinfeld: I guess not really needed since the application conf already defaults it in the same way - but might be worth putting something in here just to show where the values would go if they need to be configured like we're doing lines 36-44

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/IdentifyingAttributeCache.java
```diff
@@ -48,11 +50,15 @@ private Map<String, List<AttributeType>> loadEntityTypes(String tenantId) {
...
-            Collectors.toUnmodifiableMap(
-                EntityType::getName, this::getIdentifyingAttributesFromType));
+    try (final CloseableIterator<Document> iterator = entityTypesCollection.search(query)) {
+      return Streams.stream(iterator)
+          .flatMap(this::buildEntityType)
+          .collect(
+              Collectors.toUnmodifiableMap(
+                  EntityType::getName, this::getIdentifyingAttributesFromType));
+    } catch (final IOException e) {
+      throw new RuntimeException(e);
```
aaron-steinfeld, line 60: Can we convert this to a GRPC INTERNAL exception

entity-service-impl/src/main/java/org/hypertrace/entity/type/service/v2/EntityTypeServiceImpl.java
```diff
@@ -115,29 +117,37 @@ public void queryEntityTypes(
...
       }
+      responseObserver.onNext(
+          QueryEntityTypesResponse.newBuilder().addAllEntityType(entityTypeList).build());
+      responseObserver.onCompleted();
+    } catch (final Exception e) {
+      LOGGER.error("Unknown error occurred", e);
+      responseObserver.onError(
+          Status.INTERNAL
+              .withDescription("Unknown error occurred")
+              .asRuntimeException(requestContext.buildTrailers()));
```
aaron-steinfeld, line 149: the trailers are no longer needed as of grpc utils `0.13.1` (not sure what this repo is using). Among other things, I updated the server interceptor used for error propgation to automatically add the request context used to make the call when returning. Only need to set it in the status if we want to override that or it wasn't used to originate the call (such as from a streaming app). https://github.com/hypertrace/java-grpc-utils/blob/1f94e38d46a2ac35b73c25cb9615920386daa4cc/grpc-server-utils/src/main/java/org/hypertrace/core/grpcutils/server/ThrowableResponseInterceptor.java#L39

entity-service-impl/src/main/java/org/hypertrace/entity/fetcher/EntityFetcher.java
```diff
@@ -27,34 +30,44 @@ public EntityFetcher(
...
   }
 
-  public Stream<Entity> query(org.hypertrace.core.documentstore.Query query) {
-    return Streams.stream(this.entitiesCollection.search(query))
-        .map(this::entityFromDocument)
-        .flatMap(Optional::stream)
-        .map(Entity::toBuilder)
-        .map(Entity.Builder::build);
+  public Stream<Entity> query(org.hypertrace.core.documentstore.Query query) throws IOException {
+    final CloseableIterator<Document> iterator = this.entitiesCollection.search(query);
```
aaron-steinfeld, line 48: So we had to leave it open for the stream. Do we still need the stream or can we convert places to use the list? just .stream() it for compat if needed. If for some reason we want full compat, I'd suggest having this implemented as just `return this.query(query).stream();` Only potential downside is if there are memory reasons we're trying to avoid holding all entities in memory at one time.

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java
```diff
@@ -711,28 +784,30 @@ private <T extends Message> void searchByIdAndStreamSingleResponse(
       String entityType,
       Collection collection,
       Message.Builder builder,
-      StreamObserver<T> responseObserver) {
+      StreamObserver<T> responseObserver)
+      throws IOException {
```
aaron-steinfeld, line 788: rather than introducing a throw here and lots of new catches, since this is already responsible for sending the response to the observer (not ideal, but existing), can we just send the exception to the observer in here too. Do any catches of this method differ?

helm/templates/entity-service-config.yaml
suresh-prakash: Did some experiments earlier, but couldn't get it working. Also, repeated after this comment, but unable to make it working without special casing. @nikhil-rajesh Suggest if there are any other easy/precise ways.

helm/values.yaml
suresh-prakash: Done. Added below.

aaron-steinfeld (dismissed):

helm/templates/entity-service-config.yaml
```diff
@@ -7,20 +7,28 @@ metadata:
...
+          endpoints = {{- toJson $value }}
+              {{- else }}
           {{ $key }} = {{ $value | quote }}
+              {{- end }}
```
suresh-prakash, line 26:
```suggestion
          {{ $key }} = {{- toJson $value }}
```

nikhil-rajesh (approved):

skjindal93 (approved):

entity-service-impl/src/main/java/org/hypertrace/entity/fetcher/EntityFetcher.java
suresh-prakash: > So we had to leave it open for the stream. 

Yeah. There are 3 usages of this method, 2 of them need just the list. The third usage is actually streaming the response to the caller (1 at a time).

```java
try {
      this.entityFetcher
          .query(DocStoreConverter.transform(tenantId.get(), request, Collections.emptyList()))
          .forEach(responseObserver::onNext);
    } catch (final Exception e) {
      LOG.error("Unknown error occurred", e);
      responseObserver.onError(
          Status.INTERNAL
              .withDescription("Unknown error occurred")
              .asRuntimeException(requestContext.buildTrailers()));
    }

    responseObserver.onCompleted();
```
And, that requires us to keep the stream open.

aaron-steinfeld: In that third case - which is itself ridiculous - we can do the same with the list and close up the iterator without waiting for the client to iterate themselves. We don't want to leave the connection to the collection open at the whim of the caller.

helm/values.yaml
```diff
@@ -106,6 +102,20 @@ entityServiceConfig:
     host: attribute-service
     port: 9012
 
+database:
```
saxenakshitiz, line 105: @suresh-prakash This impacts how datastore type is configured on on-prem platform. Please keep that in mind when making such changes.

CC: @ravisingal  @kotharironak

helm/values.yaml
```diff
@@ -106,6 +102,20 @@ entityServiceConfig:
...
+  type: mongo
+  mongo:
+    authEnabled: true
+    endpoints:
+      - host: mongo
+        port: 27017
+    replicaSet: rs0
```
GurtejSohi, line 112: @suresh-prakash Why do we need to set the replica set here?

suresh-prakash: This is no longer required to be a part of the service. This has been moved to cluster-level configuration. Please refer to the continuation PR.

PR #266, opened 2023-11-25
feat: expose method to update caches with change events

kishansairam9:


entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
```diff
@@ -224,4 +226,72 @@ public void upsertEnrichedEntities(String tenantId, EnrichedEntities enrichedEnt
...
+      case CREATE_EVENT:
+        // ignore create events, don't populate caches if not necessary
+        break;
+      case UPDATE_EVENT:
+        entity = entityChangeEventValue.getUpdateEvent().getLatestEntity();
+        updateCacheValues(entityChangeEventKey, entity);
+        break;
+      case DELETE_EVENT:
+        entity = entityChangeEventValue.getDeleteEvent().getDeletedEntity();
+        entityCache.invalidate(getEntityCacheKeyWithoutEntityType(entityChangeEventKey));
```
aaron-steinfeld, line 245: doesn't this have to invalidate the one with entity type too? It seems like it would make sense to share the code that iterates the various cache entries that can be associated with an entity.

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
```diff
@@ -224,4 +226,72 @@ public void upsertEnrichedEntities(String tenantId, EnrichedEntities enrichedEnt
   public void upsertRelationships(String tenantId, EntityRelationships relationships) {
     client.upsertRelationships(tenantId, relationships);
   }
+
+  public void updateBasedOnChangeEvent(
+      EntityChangeEventKey entityChangeEventKey, EntityChangeEventValue entityChangeEventValue) {
+    LOG.debug("Entity change event is {}, {} ", entityChangeEventKey, entityChangeEventValue);
+    Entity entity;
```
aaron-steinfeld, line 233: nit - avoid the single assignment shared var. Just inline this in the function calls (e.g. `updateCacheValues(entityChangeEventKey, entityChangeEventValue.getUpdateEvent().getLatestEntity())`

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
```diff
@@ -224,4 +226,71 @@ public void upsertEnrichedEntities(String tenantId, EnrichedEntities enrichedEnt
...
+                entityChangeEventKey, entityChangeEventValue.getDeleteEvent().getDeletedEntity()));
+        break;
+      default:
+        LOG.warn(
+            "Entity change event value has invalid event type -> {}",
+            entityChangeEventValue.getEventCase());
+    }
+  }
+
+  private void updateCacheValues(EntityChangeEventKey entityChangeEventKey, Entity entity) {
```
aaron-steinfeld, line 256: I was thinking the whole method would be common between update and delete. If that seems painful to do, I would at least make update and delete consistent since they're both basically doing the same thing but one is inline in switch and the other is broken out into its own method (which I would argue is preferable)

entity-service-client/src/main/java/org/hypertrace/entity/data/service/client/EdsCacheClient.java
```diff
@@ -224,4 +226,71 @@ public void upsertEnrichedEntities(String tenantId, EnrichedEntities enrichedEnt
...
+    getEntityCacheKeys(entityChangeEventKey)
+        .forEach(
+            cacheKey -> entityCache.asMap().computeIfPresent(cacheKey, (key, oldEntity) -> entity));
```
aaron-steinfeld, line 259: sorry I missed before, but rather than converting the cache to a map and directly modifying it (two different steps that rely on understanding documented but unintuitive impl - that the map is live and modifying it modifies the cache, and that computeIfPresent has side effects that update the map) suggest using cache.put to modify the state (after doing a contains check, if needed). Among other things, I'm not sure how modify the map directly impacts things like TTLs and metrics (and whether it handles correctly or not is almost besides the point since it's not obvious one way or the other).

aaron-steinfeld (approved):

PR #270, opened 2024-01-12
Close the collection iterators

suresh-prakash:
It is a good practice to close the connections to the document-store collections. This PR closes them in all the places.


entity-service-impl/src/main/java/org/hypertrace/entity/fetcher/EntityFetcher.java
```diff
@@ -27,34 +30,44 @@ public EntityFetcher(
...
   }
 
-  public Stream<Entity> query(org.hypertrace.core.documentstore.Query query) {
-    return Streams.stream(this.entitiesCollection.search(query))
-        .map(this::entityFromDocument)
-        .flatMap(Optional::stream)
-        .map(Entity::toBuilder)
-        .map(Entity.Builder::build);
+  public Stream<Entity> query(org.hypertrace.core.documentstore.Query query) throws IOException {
+    final CloseableIterator<Document> iterator = this.entitiesCollection.search(query);
```
aaron-steinfeld, line 48: Mentioned on the other PR, suggest having this delegate to the list version. Or just changing any usage that is streaming to use the list. We don't want to leave a connection open at the whim of the ES caller. Beyond that, the iterator here is only closed if there's an exception or the stream is fully iterated. If the caller aborts mid-stream, the iterator is left open.

suresh-prakash: Yeah. I have the same feeling on why it was built that way. Entities are batched over the gRPC call in batches of 10,000. So, will it be a better idea to send multiple paginated queries to doc-store instead of directly converting the whole stream (iterator of a single query) into a list (to avoid memory spikes in the entity-service as you had mentioned before)?

aaron-steinfeld: In general, we've stopped using server side streaming everywhere, so yes, we'd ideally want to adjust callers to use a paginated approach. I'm not sure one was ever created for EDS.

I doubt we have significant concerns here on the memory - it would be good to know if any callers are really asking for more than 10k entities for example. 

That said, adding new APIs and migrating calls seems out of scope here, so if we want to be really safe and keep this behaving exactly as is let's just make sure we handle cleaning up the doc connection correctly on cancellation (once/if the caller hangs up on the stream).

suresh-prakash: Makes sense. I've updated it.

aaron-steinfeld (approved):

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java
```diff
@@ -299,21 +304,30 @@ public void delete(ByIdRequest request, StreamObserver<Empty> responseObserver)
...
+        responseObserver.onNext(Empty.newBuilder().build());
+        responseObserver.onCompleted();
+        existingEntity.ifPresent(
+            entity -> {
+              this.entityCounterMetricSender.sendEntitiesDeleteMetrics(
+                  requestContext, request.getEntityType(), List.of(entity));
+              entityChangeEventGenerator.sendDeleteNotification(requestContext, List.of(entity));
+            });
+      } else {
+        responseObserver.onError(new RuntimeException("Could not delete the entity."));
```
suresh-prakash, line 323:
```suggestion
        responseObserver.onError(new RuntimeException("Could not delete the entity"));
```

skjindal93 (approved):

aaron-steinfeld (approved):

PR #273, opened 2024-01-29
chore | adding the metrics for api count

aman-bansal:


entity-service-impl/src/main/java/org/hypertrace/entity/metric/EntityMetricsReporter.java
```diff
@@ -0,0 +1,81 @@
...
+              .config(
+                  CustomMetricConfig.builder()
+                      .metricName(ALL_API_COUNT_METRIC_NAME)
+                      .collectionName(RAW_ENTITIES_COLLECTION)
+                      .query(
+                          Query.builder()
+                              .setFilter(getFilter())
+                              .addSelection(IdentifierExpression.of(TENANT_ID_ENTITY_PATH))
+                              .addSelection(
+                                  IdentifierExpression.of(ENTITY_TYPE_ENTITY_PATH), "entity_type")
```
suresh-prakash, line 56: Can we either alias both tenant id and entity type or avoid aliasing both for consistency?
That way, all the label keys would be either in camelCase or in snake_case (but not a mix of both).

entity-service-impl/src/main/java/org/hypertrace/entity/metric/EntityMetricsReporter.java
```diff
@@ -0,0 +1,81 @@
...
+
+public class EntityMetricsReporter {
+
+  private static final String ALL_API_COUNT_METRIC_NAME = "all.api.entities.count";
+  private static final String RAW_ENTITIES_COLLECTION = "raw_entities";
+  private static final String API_DISCOVERY_STATE_ENTITY_PATH =
+      "attributes.api_discovery_state.value.string";
+  private static final String TENANT_ID_ENTITY_PATH = "tenantId";
+  private static final String ENTITY_TYPE_ENTITY_PATH = "entityType";
+  private static final String DISCOVERED = "DISCOVERED";
```
suresh-prakash, line 30: If these are coming from some enums, can we use the enums directly?

entity-service-impl/src/main/java/org/hypertrace/entity/metric/EntityMetricsReporter.java
suresh-prakash: On a separate note, given that we are already filtering by entity type API, we don't need this aggregation (since the label will always be 'API').

entity-service-impl/src/main/java/org/hypertrace/entity/metric/EntityMetricsReporter.java
```diff
@@ -0,0 +1,81 @@
...
+import org.hypertrace.core.documentstore.model.config.CustomMetricConfig;
+import org.hypertrace.core.documentstore.query.Filter;
+import org.hypertrace.core.documentstore.query.Query;
+import org.hypertrace.core.serviceframework.docstore.metrics.DocStoreCustomMetricReportingConfig;
+import org.hypertrace.core.serviceframework.docstore.metrics.DocStoreMetricsRegistry;
+import org.hypertrace.entity.v1.entitytype.EntityType;
+
+public class EntityMetricsReporter {
+
+  private static final String ALL_API_COUNT_METRIC_NAME = "all.api.entities.count";
```
suresh-prakash, line 24: Given that the individual counts (DISCOVERED, UNDER_DISCOVERY) may also be useful (and it is possible to merge metrics but not split them), can we report it per discovery state (and maybe filter out the states that are not relevant)?

```sql
SELECT tenantId, discoveryState, COUNT(1) AS value  
FROM raw_entities 
WHERE discoveryState NOT IN ["MERGED"]
GROUP BY tenantId, discoveryState
```
This way, any new discovery state introduced in the future doesn't require any changes here (unless it needs to be excluded)

suresh-prakash (commented):
Also, could you make this an incremental change on top of https://github.com/hypertrace/entity-service/pull/272/files?

entity-service-impl/src/main/java/org/hypertrace/entity/metric/EntityMetricsReporter.java
aman-bansal: no those enums are not defined in hypertrace

entity-service-impl/src/main/java/org/hypertrace/entity/metric/EntityMetricsReporter.java
```diff
@@ -1,18 +1,81 @@
...
 import org.hypertrace.core.serviceframework.docstore.metrics.DocStoreMetricsRegistry;
 import org.hypertrace.core.serviceframework.spi.PlatformServiceLifecycle;
+import org.hypertrace.entity.v1.entitytype.EntityType;
 
 public class EntityMetricsReporter {
+
+  private static final String ALL_API_COUNT_METRIC_NAME = "all.api.entities.count";
+  private static final String RAW_ENTITIES_COLLECTION = "raw_entities";
+  private static final String API_DISCOVERY_STATE_ENTITY_PATH =
+      "attributes.api_discovery_state" + ".value.string";
```
suresh-prakash, line 28: Nit: Don't think there's any significance for concatenation.
```suggestion
      "attributes.api_discovery_state.value.string";
```

entity-service-impl/src/main/java/org/hypertrace/entity/metric/EntityMetricsReporter.java
```diff
@@ -1,18 +1,81 @@
...
+import org.hypertrace.core.documentstore.query.Filter;
+import org.hypertrace.core.documentstore.query.Query;
+import org.hypertrace.core.serviceframework.docstore.metrics.DocStoreCustomMetricReportingConfig;
 import org.hypertrace.core.serviceframework.docstore.metrics.DocStoreMetricsRegistry;
 import org.hypertrace.core.serviceframework.spi.PlatformServiceLifecycle;
+import org.hypertrace.entity.v1.entitytype.EntityType;
 
 public class EntityMetricsReporter {
+
+  private static final String ALL_API_COUNT_METRIC_NAME = "all.api.entities.count";
```
suresh-prakash, line 25: Given that the metric changed now, can we now name the metric differently?
```suggestion
  private static final String API_COUNT_METRIC_NAME = "api.entities.count";
```

suresh-prakash (dismissed):

suresh-prakash (approved):

PR #282, opened 2024-05-20
fix | fix the query for fetching the updated document

aman-bansal:


skjindal93 (approved):

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java
```diff
@@ -550,6 +555,11 @@ private List<Document> getProjectedDocuments(
                     IdentifierExpression.of(EntityServiceConstants.ENTITY_ID),
                     IN,
                     ConstantExpression.ofStrings(entityIdList)))
+            .expression(
```
aaron-steinfeld, line 558: Does filter accept multiple expressions and implicitly and them or is this just overwriting the prior filter?

PR #285, opened 2024-05-22
ENG-44604: Add metrics to count no. of API entities per category per tenant

aditya-choubey2512:


entity-service-impl/src/main/java/org/hypertrace/entity/metric/EntityMetricsReporter.java
```diff
@@ -22,10 +23,13 @@
 public class EntityMetricsReporter {
 
   private static final String API_COUNT_METRIC_NAME = "api.entities.count";
+  private static final String API_CATEGORY_METRIC_NAME = "api.categories.count";
   private static final String RAW_ENTITIES_COLLECTION = "raw_entities";
   private static final String API_DISCOVERY_STATE_ENTITY_PATH =
       "attributes.api_discovery_state.value.string";
+  private static final String API_CATEGORIES_ENTITY_PATH = "attributes.categories.value";
   private static final String API_DISCOVERY_STATE = "apiDiscoveryState";
+  private static final String API_CATEGORIES = "categories";
```
suresh-prakash, line 32: Nit: Given that this will be just a single category, can we name it accordingly?
```suggestion
  private static final String API_CATEGORY = "category";
```

entity-service-impl/src/main/java/org/hypertrace/entity/metric/EntityMetricsReporter.java
```diff
@@ -22,10 +23,13 @@
 public class EntityMetricsReporter {
 
   private static final String API_COUNT_METRIC_NAME = "api.entities.count";
+  private static final String API_CATEGORY_METRIC_NAME = "api.categories.count";
   private static final String RAW_ENTITIES_COLLECTION = "raw_entities";
   private static final String API_DISCOVERY_STATE_ENTITY_PATH =
       "attributes.api_discovery_state.value.string";
+  private static final String API_CATEGORIES_ENTITY_PATH = "attributes.categories.value";
```
suresh-prakash, line 30: You may have to give a path which is an array in MongoDB. Can you double check?

suresh-prakash (approved):
Looks good to me.

skjindal93 (approved):

PR #293, opened 2024-06-14
chore | adding the support for function as an order by argument

aman-bansal:


entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/OrderByConverter.java
```diff
@@ -75,4 +89,8 @@ private static Map<org.hypertrace.entity.query.service.v1.SortOrder, SortOrder>
 
     return unmodifiableMap(map);
   }
+
+  private String getEncodedAlias(final String alias) {
```
aman-bansal, line 93: this is wrong actually. does it makes sense to encode all the sort fields? why do we encode `.` in functions but not other places?

suresh-prakash (approved):

suresh-prakash (approved):

PR #294, opened 2024-06-20
fix | fix the updates for timestamp attribute and key for update operations

aman-bansal:


entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/UpdateConverter.java
```diff
@@ -49,19 +52,21 @@
 
 @AllArgsConstructor(onConstructor_ = {@Inject})
 public class UpdateConverter implements Converter<AttributeUpdateOperation, SubDocumentUpdate> {
+
   private static final Joiner DOT_JOINER = Joiner.on(".");
 
-  private static final Map<ValueType, AttributeKind> VALUE_TYPE_TO_ATTRIBUTE_KIND_MAP =
+  private static final Map<ValueType, List<AttributeKind>> VALUE_TYPE_TO_ATTRIBUTE_KIND_MAP =
```
aaron-steinfeld, line 58: Any reason not to use multimap here?

entity-service-change-event-generator/src/main/java/org/hypertrace/entity/service/change/event/impl/EntityChangeEvaluator.java
```diff
@@ -14,9 +14,11 @@ public class EntityChangeEvaluator {
   public static ChangeResult evaluateChange(
       Collection<Entity> existingEntities, Collection<Entity> updatedEntities) {
     Map<String, Entity> existingEntityMap =
-        existingEntities.stream().collect(Collectors.toMap(Entity::getEntityId, identity()));
+        existingEntities.stream()
+            .collect(Collectors.toMap(Entity::getEntityId, identity(), (v1, v2) -> v2));
```
aaron-steinfeld, line 18: When would we have multiple entities with the same ID? isn't that a bug?

aman-bansal: not sure we are getting that from the clients. I have seen exception in upsert entities map which is defined below this. I have just added this to make it fail safe

aaron-steinfeld: My concern is that making this lenient can hide a code bug. For example, if we're updating the same entity twice then that's likely a caller bug and may indicate a problem. Rather than making this fail safe, can we investigate where we've seen that and how we're getting there?

We can of course relax this if we determine it sometimes makes sense to see the same value twice, but we should root cause if it's an upstream issue first.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/UpdateConverter.java
aman-bansal: No reason. can use multi map

entity-service-change-event-generator/src/main/java/org/hypertrace/entity/service/change/event/impl/EntityChangeEvaluator.java
aman-bansal: i have seen it with agent module entity. Let me chase that code. bit unfamiliar with that part

entity-type-service-rx-client/src/main/java/org/hypertrace/entity/type/service/rxclient/EntityTypeCachingClient.java
```diff
@@ -48,10 +44,9 @@ public Observable<EntityType> getAll() {
   }
 
   @Override
-  public Single<EntityType> get(String name) {
+  public Single<Optional<EntityType>> get(String name) {
```
aaron-steinfeld, line 47: Unnecessary breaking change. Don't change the API of a published lib. Caller can always catch so this doesn't unlock any new functionality anyway. I'm guessing you're trying to differentiate between an exception due to a missing type vs a downstream error? If so, just change the exception thrown to a not_found. Upstream, handle not found differently than other errors

entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityNormalizer.java
```diff
@@ -100,7 +103,11 @@ private boolean isV2Type(String entityType) {
     return this.entityTypeV2Client
         .get(entityType)
         .map(unused -> true)
-        .onErrorReturnItem(false)
+        .onErrorResumeNext(
+            throwable ->
+                throwable instanceof NoSuchElementException
```
aaron-steinfeld, line 108: Changing the thrown error into a grpc NOT_EXISTS exception I thought would make sense and be consistent with other places.

entity-service-impl/src/main/java/org/hypertrace/entity/query/service/converter/UpdateConverter.java
```diff
@@ -49,19 +53,23 @@
...
-          entry(STRING, TYPE_STRING),
-          entry(LONG, TYPE_INT64),
-          entry(INT, TYPE_INT64),
-          entry(FLOAT, TYPE_DOUBLE),
-          entry(DOUBLE, TYPE_DOUBLE),
-          entry(BYTES, TYPE_BYTES),
-          entry(BOOL, TYPE_BOOL),
-          entry(STRING_ARRAY, TYPE_STRING_ARRAY),
-          entry(STRING_MAP, TYPE_STRING_MAP));
+  private static final Multimap<ValueType, AttributeKind> VALUE_TYPE_TO_ATTRIBUTE_KIND_MAP =
```
aaron-steinfeld, line 59: nit - naming, now a multimap

aaron-steinfeld (approved):

